PR #29, opened 2017-06-01
Russian translation

antonv6:
It's somewhat incomplete right now, but the most obvious UI elements are translated.

There are still some hardcoded strings in settings.xml, for example:
```xml
res/xml/settings.xml:9:        android:title="Check for updates" android:entries="@array/pref_sync_frequency_titles"
res/xml/settings.xml:16:        android:title="Include mobile hotspots" android:key="pref_sync_hotspots"
res/xml/settings.xml:30:        android:title="Include mobile hotspots" android:key="pref_img_hotspots"
```

Also there are some seemingly unused strings still present (and marked as translatable), for example:
```xml
  <string name="title_section1">Section 1</string>
  <string name="title_section2">Section 2</string>
  <string name="title_section3">Section 3</string>
 ...
  <string name="dummy_button">Dummy Button</string>
  <string name="dummy_content">DUMMY\nCONTENT</string>
```
And there are strings like `<string name="hello_world">Hello world!</string>`, which are used, but are also kinda weird to see as translatable.

I'll keep an eye (or rather, a :star:) on this project, so you can merge this PR and then fix whatever you think needs fixing, and I'll catch up with the changes and update values-ru/strings.xml later.

### DIFF
--- a/app/src/main/res/values-ru/strings.xml
+++ b/app/src/main/res/values-ru/strings.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+  <string name="app_name">Feeder</string>
+  <string name="navigation_drawer_open"/>
+  <string name="navigation_drawer_close"/>
+  <string name="action_settings">Настройки</string>
+  <string name="add_feed">Добавить ленту</string>
+  <string name="title_activity_edit_feed">Подписаться на ленту</string>
+  <string name="url">URL</string>
+  <string name="tag">Тег</string>
+  <string name="add_feed_search_hint">URL ленты</string>
+  <string name="title">Название</string>
+
+  <string name="save">Сохранить</string>
+  <string name="empty_feed_top">Здесь нечего читать. Хотите…</string>
+  <string name="empty_feed_open">
+ &lt;font color=#607D8B&gt;Открыть&lt;/font&gt; другую ленту?</string>
+  <string name="empty_feed_add">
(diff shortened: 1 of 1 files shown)


spacecowboy:
@antonv6 Thanks for the translation!

Yes there a couple of strings lying around which are only used in the layout editor and never shown in actual use.

I'll make sure the settings strings are translatable!

(oh and FYI this is just a mirror to https://gitlab.com/spacecowboy/Feeder . Merging the PR is no problem, but I think the mirror process runs once a day or so, in case of lag)

app/src/main/res/values-ru/strings.xml
```diff
@@ -0,0 +1,54 @@
...
+  <string name="title_activity_edit_feed">Подписаться на ленту</string>
+  <string name="url">URL</string>
+  <string name="tag">Тег</string>
+  <string name="add_feed_search_hint">URL ленты</string>
+  <string name="title">Название</string>
+
+  <string name="save">Сохранить</string>
+  <string name="empty_feed_top">Здесь нечего читать. Хотите…</string>
+  <string name="empty_feed_open">
+ &lt;font color=#607D8B&gt;Открыть&lt;/font&gt; другую ленту?</string>
```
spacecowboy: These strings will show html code to the user.

Compare with the original: https://gitlab.com/spacecowboy/Feeder/blob/master/app/src/main/res/values/strings.xml#L31

```
<string name="empty_feed_open">
    <![CDATA[<font color="#607D8B">Open</font> another feed?]]></string>
```

PR #31, opened 2017-07-19
Update Russian translation

antonv6:
Not sure if `nothing_to_display` is used anywhere right now, but it also got translated.


spacecowboy:
Thanks for the PR! Will merge shortly:
https://gitlab.com/spacecowboy/Feeder/merge_requests/48

PR #34, opened 2017-09-14
Update strings.xml

spacecowboy:
@marcoM32 I opened a new PR for these changes (makes it easier to handle the merge)


spacecowboy:
@marcoM32 btw, there doesn't appear to be any actual changes in the translation here (and you've replaced the `...` character with 3 literal dots). Or am I missing something?

(Note that the `CDATA` stuff is equivalent to the escaped html I inserted)

marcoM32:
Hi @spacecowboy, the only change is in CDATA string if this is not needed this pull request is useless. The dots are a regression and shoul be fixed.

spacecowboy:
OK then I'll close it. But thanks again for the translation!

PR #50, opened 2023-05-13
Dim Text for Articles Marked as Read

n8x0b7T:
Finding it difficult to distinguish read vs unread articles in the app, I though it would be good to dim article text for read items. In these commits, I choose an opacity value based on if the article was marked as read. I then set the alpha value for each of the text elements for every article style. This approach works well for both light and dark themes.


spacecowboy:
Thanks for the PR but I think the difference is good as is.

PR #51, opened 2023-05-15
Set jvm target to 11

syphyr:
Fixes the following:

> Task :jsonfeed-parser:compileKotlin
'compileJava' task (current target is 11) and 'compileKotlin' task
  (current target is 1.8) jvm target compatibility should be set to
  the same Java version.
By default will become an error since Gradle 8.0+! Read more:
  https://kotl.in/gradle/jvm/target-validation
Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -44,6 +44,10 @@ android {
                 argument("room.incremental", "true")
             }
         }
+
+        kotlinOptions {
+            jvmTarget = "11"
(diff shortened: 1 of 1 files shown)


spacecowboy:
Thanks but please don't submit PRs for build scripts.

Also, there is already a `kotlinOptions` block in the file which you missed

PR #64, opened 2023-06-06
Improved keyboard navigation through the app

spacecowboy:


a1batross:
Hey @spacecowboy you seem disabled issues everyone. How does one should send bugrepors?

PR #105, opened 2023-10-03
Add ability to search current feed by title of feed item

omacranger:
Hey! Love the app. Use it all the time. I've ran into this a few times, where I've seen an article then it's been a few hours and trying to find it again proves really difficult with a few hundred items in the list. I've added a search icon on the navigation bar which converts to a search field and filters the current page query by the text entered. It's my first time using Compose and Kotlin so bear with me, but I think I followed the repo's expectations fairly well.

Only concern I have with the 'functionality' is filtering a list then using one of the "Mark above as read...", etc, marks any newer / older items as read instead of the filtered ones. It wouldn't be terribly difficult to go back and have those updated with that same behavior, but figured I would ask here before I did that as well (and if the whole PR would be accepted in general).

Example here:

https://github.com/spacecowboy/Feeder/assets/3063830/ef3ec6e0-2ebe-4a4c-9de5-23b147ea6e61

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
@@ -65,6 +65,7 @@ class FeedItemStore(override val di: DI) : DIAware {
         minReadTime: Instant,
         newestFirst: Boolean,
         filter: FeedListFilter,
+        searchText: String?,
     ): Flow<PagingData<FeedListItem>> =
         Pager(
             config = PagingConfig(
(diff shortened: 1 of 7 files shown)


spacecowboy:
Cool that you wanted to tackle this.

Unfortunately I'm not willing to accept it in its current form.

A search bar should conform to the material design guidelines for a `Search view` here:

https://m3.material.io/components/search/guidelines#3f0ecf2a-cc1e-48de-863b-0adb0be3c03b

and special consideration needs to be taken to have the correct behavior also for large screens like tablets.

I have begun some work to implement that in redesigned Add Feed screen which can be viewed (very much early testing here!) here:
https://github.com/spacecowboy/Feeder/blob/new-add-screen/app/src/main/java/com/nononsenseapps/feeder/ui/compose/searchfeed/AddFeedScreen.kt

I understand if this is discouraging, since you already have a solution which works for you. But let me know if you'd be willing to tackle it this way. I'd be happy to actively assist with the work. Feedback and/or code.

omacranger:
Makes sense! Appreciate the feedback. If they have some of those search view components / overlay available in the material library I'll take a look and see what I can do, otherwise it may be out of my capabilities.

PR #108, opened 2023-10-19
Fix notifications not following settings for item opening

NobodyForNothing:
### Motivation
The problem this attempts to fix is that when clicking on a single item notification, the app is opened, ignoring the settings for opening items.
While a dedicated action for opening in the browser is present, it does not solve the issue, as an additional click is required to expand the buttons. 

### Implemented changes
This PR opens the item in the browser directly and hides the "open in browser" button when the item's default is set this way.
image

### Potential issues with this PR
Notification builders only check for opening in the browser there might be a more general way to implement this. If this is a concern please point out the functions needed so I can modify the code accordingly.

The "open in browser" button could be replaced with an "open in app" button when the browser setting is set. I did not implement this to keep the changes simple.

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/model/RssNotifications.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/model/RssNotifications.kt
@@ -125,7 +125,7 @@ private fun createNotificationChannel(context: Context) {
     notificationManager.createNotificationChannel(channel)
 }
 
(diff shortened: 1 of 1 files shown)


spacecowboy:
@NobodyForNothing Thanks for the PR! It looks good to me, I've told github to merge once the pipeline succeeds.

spacecowboy (approved):

PR #134, opened 2023-11-23
Added reading time/word count for languages which use spaces

spacecowboy:


BoostCookie:
Now it always shows me the reading time, even when I have it disabled in the settings.

spacecowboy:
> Now it always shows me the reading time, even when I have it disabled in the settings.

The setting is for the list items as you can see by the section name

PR #147, opened 2023-12-02
Improved image rendered and scrolling

spacecowboy:


Fs00:
Oh, I see some nice work here! Looking forward to the improved reader scrolling smoothness 👀
Since you are working on images rendering, I'd like to point out two issues I've noticed:
- if image download is enabled only on Wi-Fi networks, an error icon is rendered in place of images when I'm connected to the mobile network. Is that the expected behaviour? I'd rather expect images not to be present at all or to be replaced by a neutral placeholder.
- images in xda-developers.com articles fail to be rendered. I'm assuming that's because the image URL is placed in the `data-img-url` attribute of `` tags (you can see it in [this article](https://www.xda-developers.com/microsoft-onedrive-in-teams/) for example). Would it be possible to properly handle those images?

spacecowboy:
>     * if image download is enabled only on Wi-Fi networks, an error icon is rendered in place of images when I'm connected to the mobile network. Is that the expected behaviour? I'd rather expect images not to be present at all or to be replaced by a neutral placeholder.

Expected behavior. It is implemented in such a way that an error is thrown when no network is available and the image is not cached.

>     * images in xda-developers.com articles fail to be rendered. I'm assuming that's because the image URL is placed in the `data-img-url` attribute of `` tags (you can see it in [this article](https://www.xda-developers.com/microsoft-onedrive-in-teams/) for example). Would it be possible to properly handle those images?

Added

PR #185, opened 2024-01-11
Implemented 'Exclude from "All feeds"' setting

c-andy-candies:
Someone requested it in discussion #122 and I recently started having the same problem.

This commit adds a feed setting "Exclude from All feeds", if set the feed items of that feed won't appear in the All feeds tab.

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
@@ -134,6 +134,7 @@ class FeedItemStore(override val di: DI) : DIAware {
             onlySavedArticles -> append("AND bookmarked = 1\n")
             feedId > ID_UNSET -> append("AND feed_id IS ?\n").also { args.add(feedId) }
             tag.isNotEmpty() -> append("AND tag IS ?\n").also { args.add(tag) }
+            feedId <= ID_UNSET -> append("AND (SELECT exclude_from_all_feeds FROM feeds WHERE feeds.id = feed_items.feed_id) IS 0\n")
         }
     }
 
--- a/app/src/main/java/com/nononsenseapps/feeder/db/Constants.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/db/Constants.kt
@@ -50,6 +50,7 @@ const val COL_READ_TIME = "read_time"
 const val COL_SITE_FETCHED = "site_fetched"
 const val COL_WORD_COUNT = "word_count"
 const val COL_WORD_COUNT_FULL = "word_count_full"
+const val COL_EXCLUDE_FROM_ALL_FEEDS = "exclude_from_all_feeds"
 
 // year 5000
 val FAR_FUTURE = Instant.ofEpochSecond(95635369646)
--- a/app/src/main/java/com/nononsenseapps/feeder/db/room/AppDatabase.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/db/room/AppDatabase.kt
@@ -127,12 +127,23 @@ fun getAllMigrations(di: DI) =
         MigrationFrom29To30(di),
         MigrationFrom30To31(di),
         MigrationFrom31To32(di),
+        MigrationFrom32To33(di),
     )
 
 /*
  * 6 represents legacy database
  * 7 represents new Room database
  */
+class MigrationFrom32To33(override val di: DI) : Migration(32, 33), DIAware {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(
+            """
+            alter table feeds add column exclude_from_all_feeds integer not null default 0
+            """.trimIndent(),
+        )
+    }
+}
+
 class MigrationFrom31To32(override val di: DI) : Migration(31, 32), DIAware {
     override fun migrate(database: SupportSQLiteDatabase) {
         database.execSQL(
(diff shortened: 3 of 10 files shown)


spacecowboy:
Thanks but no thanks. The all feeds is what it's called, All feeds. I see no reason why feeds should be excluded from it.

The discussion you are referring to was about adding multiple tags support to feeds.

c-andy-candies:
That was part of the discussion about workarounds for what the original poster was feature requesting: since you and others told him that what he wanted to do was not possible, he was asking instead about multiple tags. In fact he said about using tags:

> although it doesn't exactly solve my problem, thanks for the idea.

My use case and the one from the original poster of #122 are similar: I use All feeds basically all the time and then sometimes want to see specific tags or specific feeds, unfortunately there are sources that multiple times a day have a huge number of feed items and that buries all other items. I don't want to stop receiving updates from those flooding feeds but I want to do it at my pace, contrary to all other feeds which are updating at a less frequent pace. 

You said you don't use the "All feeds" tab, then why have it in the first place? Since tags are not compulsory for a feed I'm sure many people, including me, are using it and this tries to improve on this way of using the app.

The problem with the suggested workaround is that if someone doesn't want to tag all feeds and categorize them, they would end up with two categories: "Fake All feeds" and "Not included in All feeds" and a third useless "All feeds".

Hope I provided some reasoning for this PR. Please don't penalize people with a workflow different from yours.

The app is great, thank you for it!

spacecowboy:
First let me acknowledge that it's great that you want to contribute to the app. I don't want discourage that. 

Major changes are best to discuss with me before doing a bunch of work on them. They need to fit into the bigger picture and potential future work as well.

> My use case and the one from the original poster of #122 are similar: I use All feeds basically all the time and then sometimes want to see specific tags or specific feeds, unfortunately there are sources that multiple times a day have a huge number of feed items and that buries all other items. I don't want to stop receiving updates from those flooding feeds but I want to do it at my pace, contrary to all other feeds which are updating at a less frequent pace.

I problem I can sympathize with as I deal with that myself.

> You said you don't use the "All feeds" tab, then why have it in the first place? Since tags are not compulsory for a feed I'm sure many people, including me, are using it and this tries to improve on this way of using the app.

Yes it tries to improve it, but I think it's more of a workaround than a real solution. I'd prefer to discuss possibilities for a real improvement instead.

One obvious objection I have is that if you can exclude feeds from the "All feeds" meta-tag, then it can't be called "All feeds" anymore.

> The problem with the suggested workaround is that if someone doesn't want to tag all feeds and categorize them, they would end up with two categories: "Fake All feeds" and "Not included in All feeds" and a third useless "All feeds".

Sure, but that would solve the problem within the existing constraints though. I'm not a fan of adding complexity just to have a second way of solving something which can already be solved in the app.

Instead, let's discuss how to solve the issue properly and see if you want to be a part of that.

One thing I've considered is adding more "meta-tags" like "All Feeds" (or replacing it).

Consider for example what NetNewsWire calls "Smart Feeds" in this image:

Example from NetNewsWire

Where you have 

* Today
* All Unread
* Starred

One could imagine others, and the ability to exclude feeds from them (since none are called All Feeds).

c-andy-candies:
> First let me acknowledge that it's great that you want to contribute to the app. I don't want discourage that.
> 
> Major changes are best to discuss with me before doing a bunch of work on them. They need to fit into the bigger picture and potential future work as well.

Would you consider adding a CONTRIBUTING.md to the repo with contributing guidelines and all the infos necessary to contribute? I thought you didn't accept contributions when I saw there was none.

> One obvious objection I have is that if you can exclude feeds from the "All feeds" meta-tag, then it can't be called "All feeds" anymore.

If the name is the problem, we can think about it: "Frontpage", "Main feed"?

> Instead, let's discuss how to solve the issue properly and see if you want to be a part of that.

I will participate if I'll be able to, timewise, skillwise, etc.

> One could imagine others, and the ability to exclude feeds from them (since none are called All Feeds).

I think these would be a very good addition to the app for organizing feeds, but I feel  that the way of using the app I described would still be weird to achieve since I would need a "Mix all feeds except X" meta-tag and a "X" meta-tag. 

I'm not saying I can't / won't change the way I use the app to adapt to it, I'm just saying the status quo and the meta-tag proposal "force" you to organize the feeds.
It isn't a bad thing per se, but if you just want to randomly scroll through feeds from multiple mixed and unorganized feeds, which is the spirit of the already present All feeds meta-tag (even though that spirit clashes against the reality of flooding feed sources), you have to adopt a workaround.

spacecowboy:
> I'm not saying I can't / won't change the way I use the app to adapt to it, I'm just saying the status quo and the meta-tag proposal "force" you to organize the feeds. It isn't a bad thing per se, but if you just want to randomly scroll through feeds from multiple mixed and unorganized feeds, which is the spirit of the already present All feeds meta-tag (even though that spirit clashes against the reality of flooding feed sources), you have to adopt a workaround.

The underlying problem is that some feeds are spammy, right?

Well what about a "Just give me stuff to read" meta tag which does not show simply chronologically, but uses some logic to prevent a single feed to dominate the list? That way you don't even need to exclude them.

KAGEYAM4:
What about option create custom feeds section, usesr can choose the name ( eg. Fitelerd )

spacecowboy:
> What about option create custom feeds section, usesr can choose the name ( eg. Filtered )

what do you mean by section?

KAGEYAM4:
> > What about option create custom feeds section, usesr can choose the name ( eg. Filtered )
> 
> what do you mean by section?

Folder/Section - In the left hamburger menu

Some rules can be set for folder/Section - where ->

- Exlude certain rss feeds
- Only include certain rss feeds
- Maybe in future more custom operation for filtering the feeds in a folder/section

anaxonda:
Can this feature be looked at again? Currently there is the 'block' setting that matches titles to exclude. I would love to be able to block feeds matching keywords also.

PR #234, opened 2024-03-11
feat: add select all button to delete feed

dracarys18:
- Added a Select All button in deleted feed screen
- Formatting

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
@@ -138,7 +138,7 @@ fun DeleteFeedDialog(
                         }, onClick = {
                             onSelectAll(!selectAllSelected)
                             menuExpanded = false
-                        }, modifier = Modifier.safeSemantics(mergeDescendants = selectAllSelected) {
+                        }, modifier = Modifier.safeSemantics(mergeDescendants = true) {
                             stateDescription = stateLabel
                         })
                     }
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
@@ -56,6 +56,9 @@ fun DeleteFeedDialog(
         isChecked = { feedId ->
             feedsToDelete[feedId] ?: false
         },
+        isSelectAllChecked = {
+            if (feedsToDelete.isEmpty()) false else feedsToDelete.all { it.value }
+        },
         onDismiss = onDismiss,
         onOk = {
             onDelete(feedsToDelete.filterValues { it }.keys)
@@ -77,6 +80,7 @@ fun DeleteFeedDialog(
 fun DeleteFeedDialog(
     feeds: ImmutableHolder<List<DeletableFeed>>,
     isChecked: (Long) -> Boolean,
+    isSelectAllChecked: () -> Boolean,
     onDismiss: () -> Unit,
     onOk: () -> Unit,
     onToggleFeed: (Long, Boolean?) -> Unit,
@@ -121,11 +125,13 @@ fun DeleteFeedDialog(
 
                     DropdownMenu(expanded = menuExpanded, onDismissRequest = { menuExpanded = false }) {
                         DropdownMenuItem(text = {
-                            Row(modifier = Modifier.clickable {
-                                onSelectAll(!selectAllSelected)
-                                selectAllSelected = !selectAllSelected
-                                menuExpanded = false
-                            }, verticalAlignment = Alignment.CenterVertically) {
+                            Row(verticalAlignment = Alignment.CenterVertically) {
+                                Checkbox(checked = isSelectAllChecked(), onCheckedChange = { checked ->
+                                    onSelectAll(checked)
+                                    selectAllSelected = checked
+                                    menuExpanded = false
+                                }, modifier = Modifier.clearAndSetSemantics { })
+                                Spacer(modifier = Modifier.width(4.dp))
                                 Text(
                                     text = stringResource(id = android.R.string.selectAll),
                                     style = MaterialTheme.typography.titleMedium,
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
@@ -56,6 +56,9 @@ fun DeleteFeedDialog(
         isChecked = { feedId ->
             feedsToDelete[feedId] ?: false
         },
+        isSelectAllChecked = {
+            if (feedsToDelete.isEmpty()) false else feedsToDelete.all { it.value }
+        },
         onDismiss = onDismiss,
         onOk = {
             onDelete(feedsToDelete.filterValues { it }.keys)
@@ -66,7 +69,7 @@ fun DeleteFeedDialog(
         },
         onSelectAll = { checked ->
             feeds.item.forEach { item ->
-                feedsToDelete = feedsToDelete + (item.id to (checked ?: !feedsToDelete.contains(item.id)))
+                feedsToDelete = feeds.item.associate { item.id to checked }
             }
         },
         modifier = modifier,
@@ -77,16 +80,17 @@ fun DeleteFeedDialog(
 fun DeleteFeedDialog(
     feeds: ImmutableHolder<List<DeletableFeed>>,
     isChecked: (Long) -> Boolean,
+    isSelectAllChecked: () -> Boolean,
     onDismiss: () -> Unit,
     onOk: () -> Unit,
     onToggleFeed: (Long, Boolean?) -> Unit,
-    onSelectAll: (Boolean?) -> Unit,
+    onSelectAll: (Boolean) -> Unit,
     modifier: Modifier = Modifier,
 ) {
     var selectAllSelected by rememberSaveable {
         mutableStateOf(false)
     }
-    var selectAllButton by rememberSaveable {
+    var menuExpanded by rememberSaveable {
         mutableStateOf(false)
     }
     AlertDialog(
@@ -115,16 +119,17 @@ fun DeleteFeedDialog(
                     modifier = Modifier.align(Alignment.CenterStart)
                 )
                 Box(modifier = Modifier.align(Alignment.CenterEnd)) {
-                    IconButton(onClick = { selectAllButton = true }) {
-                        Icon(Icons.Default.MoreVert, contentDescription = "Select All")
+                    IconButton(onClick = { menuExpanded = true }) {
+                        Icon(Icons.Default.MoreVert, contentDescription = stringResource(android.R.string.selectAll))
                     }
 
-                    DropdownMenu(expanded = selectAllButton, onDismissRequest = { selectAllButton = false }) {
+                    DropdownMenu(expanded = menuExpanded, onDismissRequest = { menuExpanded = false }) {
                         DropdownMenuItem(text = {
                             Row(verticalAlignment = Alignment.CenterVertically) {
-                                Checkbox(checked = selectAllSelected, onCheckedChange = { checked ->
+                                Checkbox(checked = isSelectAllChecked(), onCheckedChange = { checked ->
                                     onSelectAll(checked)
                                     selectAllSelected = checked
+                                    menuExpanded = false
                                 }, modifier = Modifier.clearAndSetSemantics { })
                                 Spacer(modifier = Modifier.width(4.dp))
                                 Text(
@@ -135,6 +140,7 @@ fun DeleteFeedDialog(
                         }, onClick = {
                             onSelectAll(!selectAllSelected)
                             selectAllSelected = !selectAllSelected
+                            menuExpanded = false
                         }, modifier = Modifier.safeSemantics(mergeDescendants = selectAllSelected) {
                             stateDescription = stateLabel
                         })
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -86,14 +103,44 @@ fun DeleteFeedDialog(
...
+                )
+                Box(modifier = Modifier.align(Alignment.CenterEnd)) {
+                    IconButton(onClick = { selectAllButton = true }) {
+                        Icon(Icons.Default.MoreVert, contentDescription = "Select All")
+                    }
+
+                    DropdownMenu(expanded = selectAllButton, onDismissRequest = { selectAllButton = false }) {
+                        DropdownMenuItem(text = {
+                            Row(verticalAlignment = Alignment.CenterVertically) {
+                                Checkbox(checked = selectAllSelected, onCheckedChange = { checked ->
```
spacecowboy, line 125: IMO this checkbox can be removed. It's just extra logic to keep that checkbox in a correct state. What do you think?

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -59,6 +64,11 @@ fun DeleteFeedDialog(
         onToggleFeed = { feedId, checked ->
             feedsToDelete = feedsToDelete + (feedId to (checked ?: !feedsToDelete.contains(feedId)))
         },
+        onSelectAll = { checked ->
+            feeds.item.forEach { item ->
+                feedsToDelete = feedsToDelete + (item.id to (checked ?: !feedsToDelete.contains(item.id)))
+            }
```
spacecowboy, line 70:
```suggestion
            feedsToDelete = feeds.item.associate { it.id to true }
```

Can be simplified

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -70,8 +80,15 @@ fun DeleteFeedDialog(
     onDismiss: () -> Unit,
     onOk: () -> Unit,
     onToggleFeed: (Long, Boolean?) -> Unit,
+    onSelectAll: (Boolean?) -> Unit,
```
spacecowboy, line 83:
```suggestion
    onSelectAll: (Boolean) -> Unit,
```

spacecowboy: No nulls

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -70,8 +80,15 @@ fun DeleteFeedDialog(
...
     modifier: Modifier = Modifier,
 ) {
+    var selectAllSelected by rememberSaveable {
+        mutableStateOf(false)
+    }
+    var selectAllButton by rememberSaveable {
```
spacecowboy, line 89:
```suggestion
    var menuExpanded by rememberSaveable {
```

spacecowboy (changes requested):
Thanks for the PR! It looks pretty good.

I've noted some comments and suggestions.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
dracarys18: What do you prefer instead of checkbox? A normal button?

spacecowboy: no just let it be text in that case. most dropdowns in android are just plain text

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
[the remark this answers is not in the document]
dracarys18: I have addressed this

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -70,8 +80,15 @@ fun DeleteFeedDialog(
     onDismiss: () -> Unit,
     onOk: () -> Unit,
     onToggleFeed: (Long, Boolean?) -> Unit,
+    onSelectAll: (Boolean?) -> Unit,
     modifier: Modifier = Modifier,
 ) {
+    var selectAllSelected by rememberSaveable {
```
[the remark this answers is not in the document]
dracarys18: Addressed this in the new commit

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
dracarys18: Alright will update

dracarys18: Remove the checkbox, But I personally would prefer it tho, It would give me more insights if all the values are selected or not

dracarys18:
> You could keep it simple (because it's tricky) and just make the select-all button be dumb: it always selects all.

I would prefer it to be a checkbox with deselect. Because let's say I selected all of the entries with Select All button and now I have to deselect them then I have to do it one by one which is not really convenient. What are your thoughts?

spacecowboy:
> I would prefer it to be a checkbox with deselect. Because let's say I selected all of the entries with Select All button and now I have to deselect them then I have to do it one by one which is not really convenient. What are your thoughts?

Sure, I don't mind. But it's tricky as I said.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -51,6 +56,9 @@ fun DeleteFeedDialog(
         isChecked = { feedId ->
             feedsToDelete[feedId] ?: false
         },
+        isSelectAllChecked = {
+            if (feedsToDelete.isEmpty()) false else feedsToDelete.all { it.value }
```
spacecowboy, line 60: This actually doesn't work.

If one manually adds items to to delete, then the second case will return true because `feedsToDelete` will contain a sub-set of all items that are selectable.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -86,14 +107,46 @@ fun DeleteFeedDialog(
...
+                )
+                Box(modifier = Modifier.align(Alignment.CenterEnd)) {
+                    IconButton(onClick = { menuExpanded = true }) {
+                        Icon(Icons.Default.MoreVert, contentDescription = stringResource(android.R.string.selectAll))
+                    }
+
+                    DropdownMenu(expanded = menuExpanded, onDismissRequest = { menuExpanded = false }) {
+                        DropdownMenuItem(text = {
+                            Row(verticalAlignment = Alignment.CenterVertically) {
+                                Checkbox(checked = isSelectAllChecked(), onCheckedChange = { checked ->
```
spacecowboy, line 129:
```suggestion
                                Checkbox(checked = isSelectAllChecked, onCheckedChange = { checked ->
```

Don't execute the lambda

spacecowboy (changes requested):

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
dracarys18: the `all` function will return true only if all the values here are true.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
dracarys18: `Checked` here expects a Boolean, How would you prefer this?

spacecowboy: Doh..

Well, in that case it can't be a lambda. That won't work very well I think. Needs to be a state.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
spacecowboy: Yes, but the `feedsToDelete` map starts empty. Then you get items added as they are checked.

Consider 10 items.

First the map is empty
Then it has one item - and `.all { ... }` will return true, which will be incorrect.

You need to check if all of `feeds.items` are true. Something like

```
feeds.items.all { feedsToDelete[it.id] }
```

spacecowboy: Tried it on a local checkout and you can change the lambda to this

```
val isSelectAllChecked by remember {
        derivedStateOf {
            feeds.item.all { feed -> feedsToDelete[feed.id] ?: false }
        }
    }
```

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -59,6 +67,11 @@ fun DeleteFeedDialog(
         onToggleFeed = { feedId, checked ->
             feedsToDelete = feedsToDelete + (feedId to (checked ?: !feedsToDelete.contains(feedId)))
         },
+        onSelectAll = { checked ->
+            feeds.item.forEach { item ->
+                feedsToDelete = feeds.item.associate { item.id to checked }
+            }
```
spacecowboy, line 73:
```suggestion
         onSelectAll = { checked ->
            feedsToDelete = feeds.item.associate { it.id to checked }
        },
```

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -67,11 +80,19 @@ fun DeleteFeedDialog(
 fun DeleteFeedDialog(
     feeds: ImmutableHolder<List<DeletableFeed>>,
     isChecked: (Long) -> Boolean,
+    isSelectAllChecked: () -> Boolean,
```
spacecowboy, line 83:
```suggestion
    isSelectAllChecked: Boolean,
```

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -67,11 +80,19 @@ fun DeleteFeedDialog(
...
     onDismiss: () -> Unit,
     onOk: () -> Unit,
     onToggleFeed: (Long, Boolean?) -> Unit,
+    onSelectAll: (Boolean) -> Unit,
     modifier: Modifier = Modifier,
 ) {
+    var selectAllSelected by rememberSaveable {
+        mutableStateOf(false)
+    }
```
spacecowboy, line 92:
```suggestion

```

Replaced by outer state

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -86,14 +107,46 @@ fun DeleteFeedDialog(
...
-            Text(
-                text = stringResource(id = R.string.delete_feed),
-                style = MaterialTheme.typography.titleLarge,
-                textAlign = TextAlign.Center,
-                modifier =
-                    Modifier
-                        .padding(vertical = 8.dp),
-            )
+            Box(modifier = Modifier.fillMaxWidth()) {
+                val stateLabel = if (selectAllSelected) {
```
spacecowboy, line 111:
```suggestion
                val stateLabel = if (isSelectAllChecked) {
```

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
spacecowboy, line 129:
```suggestion
                                Checkbox(checked = isSelectAllChecked, onCheckedChange = { checked ->
```

app/src/main/java/com/nononsenseapps/feeder/ui/compose/deletefeed/DeleteFeedScreen.kt
```diff
@@ -86,14 +107,46 @@ fun DeleteFeedDialog(
...
+                                Text(
+                                    text = stringResource(id = android.R.string.selectAll),
+                                    style = MaterialTheme.typography.titleMedium,
+                                )
+                            }
+                        }, onClick = {
+                            onSelectAll(!selectAllSelected)
+                            selectAllSelected = !selectAllSelected
+                            menuExpanded = false
+                        }, modifier = Modifier.safeSemantics(mergeDescendants = selectAllSelected) {
```
spacecowboy, line 144:
```suggestion
                        }, modifier = Modifier.safeSemantics(mergeDescendants = true) {
```

spacecowboy (changes requested):
Some additional changes after adding the `derivedState` mentioned previously

dracarys18:
> Some additional changes after adding the `derivedState` mentioned previously

I have resolved them please check

spacecowboy (approved):

spacecowboy:
Looks good now @dracarys18 Thanks for your contribution!

PR #243, opened 2024-03-15
Proposal - compact card layout

anod:
This proposal PR adds another style for cards - compact cards, see screenshot
screen

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -443,6 +443,6 @@ dependencyResolutionManagement {
     }
 }
 
-rootProject.name = "Feeder"
+rootProject.name = "feeder"
 
 include(":app")
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -443,6 +443,6 @@ dependencyResolutionManagement {
     }
 }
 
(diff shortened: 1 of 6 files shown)


settings.gradle.kts
```diff
@@ -443,6 +443,6 @@ dependencyResolutionManagement {
     }
 }
 
-rootProject.name = "feeder"
+rootProject.name = "Feeder"
```
spacecowboy, line 446: This doesn't seem relevant to this PR.

Rename your directory to `feeder` instead?

Maybe I can rename the repo or something..

spacecowboy (changes requested):
This looks good to me, except for the second commit (see comment). Just drop that and it can be merged IMO.

settings.gradle.kts
anod: Renaming the folder to lowercase didn't work. The app won't sync for me without it. will drop for this change, but I need to apply it for myself to build the app

image)

spacecowboy (approved):

spacecowboy:
Thanks again @anod ! Great work

PR #347, opened 2024-08-08
Fix the scaling of small thumbnails

ymcx:
Small images appear to not scale correctly when Precision.INEXACT is used rather than Precision.EXACT. As you can see from the images below, with the fix applied the thumbnails scale to fit the frame rather than having the huge border shown.
Please note that the feed is somewhat broken and their thumbnails are low quality and sometimes stretched.

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCard.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCard.kt
@@ -125,7 +125,7 @@ fun FeedItemCard(
                                         )
                                         .scale(Scale.FIT)
                                         .size(pixels)
-                                        .precision(Precision.INEXACT)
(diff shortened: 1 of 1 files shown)


spacecowboy:
The border is actually intentional. Personally, I preferred to see the images small and clear instead of large and pixly.

Perhaps it could be a setting if you agree?

ymcx:
> The border is actually intentional. Personally, I preferred to see the images small and clear instead of large and pixly.
> 
> Perhaps it could be a setting if you agree?

The 'compact card' layout seems to hande thumbnails by resizing them to the image boundaries, which is why I initially thought the current approach wasn't intentional.

However, I totally see your point about not wanting them to become blurry since like in my example, your feed starts getting really ugly if it's filled with entries with low resolution images.

A setting to either
a) hide all of the thumbnails in a given feed or
b) force the thumbnails to the size of the frame like in the initial commit or maybe even
c) both
would probably be sufficient.

PR #352, opened 2024-08-19
Added comparison between read and unread in Compact and Super Compact modes.

bannedbook:
Added comparison between read and unread article titles in Compact and Super Compact modes.

In Compact and Super Compact modes, it is difficult to distinguish between read and unread articles. This PR will improve this problem.

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCard.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCard.kt
@@ -67,7 +67,8 @@ import com.nononsenseapps.feeder.ui.compose.utils.PreviewThemes
 import com.nononsenseapps.feeder.ui.compose.utils.onKeyEventLikeEscape
 import java.net.URL
 import java.time.Instant
-
+import androidx.compose.ui.graphics.Color
+import androidx.compose.material3.LocalContentColor
 @Composable
 fun FeedItemCard(
     item: FeedListItem,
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCard.kt
```diff
@@ -210,6 +211,12 @@ fun RowScope.FeedItemText(
             }
         }
     val closeMenuText = stringResource(id = R.string.close_menu)
+    // The color of unread items remains unchanged, and the color of read items increases transparency
+    val titleColor = if (item.unread) {
+        Color.Unspecified // Use the default color and do not change the color of unread items
+    } else {
+        LocalContentColor.current.copy(alpha = 0.74f) //Increase transparency of read item color
+    }
```
spacecowboy, line 219: All of this should be moved inside the `FeedListItemTitleTextStyle` function in `Typography.kt`. See how the the font weight uses the unread flag:

```
fontWeight = titleFontWeight(item.unread)
```

spacecowboy (changes requested):
See inline comment.

Also by `comparison` you actually mean `difference`. And this change will affect all styles, not only compact/super compact.

spacecowboy:
Also please run `./gradlew ktlintFormat` on the code.

bannedbook:
Yes, yes, this change will affect all styles. 
But before this, In Compact and Super Compact modes, it is difficult to distinguish between read and unread articles. 
So, it mainly solves the problem of Compact and Super Compact, but it also affects other modes, but it does not destroy the effect, which should have a better effect  , but the effect is just not obvious .

PR #386, opened 2024-10-12
Open drawer after hitting fab mark all articles

### LINKED ISSUE #375: Open drawer after hitting fab button to mark all articles read
### Checklist

### App version

2.6.32

Other

It would a great QOL improvment if when hitting fab button to mark all read then drawer was opened. You could then choose another tag to continue reading.

### Why should this be added?

When you mark all articles read you end up on an empty screen where there is nothing to do. To continue reading articles from others tags the only choice is to manually open the drawer.
So it would be nice for users if it was opening without having to do it manually to continue reading or to check if other tags with unread article exists.

_No response_

maksimowiczm:
From now hitting mark all articles fab will open drawer.

Closes  #375


spacecowboy:
Thanks for the PR @maksimowiczm !

PR #388, opened 2024-10-13
fix: feed title empty on initial app start

svenjacobs:
On the first app start the feed title is empty.

Screenshot_20241013_124246

`SettingsStore` returns `ID_UNSET` for the current feed because the user has not explicitly selected a feed yet.

The PR fixes it by setting `FeedType` to `ALL_FEEDS` when ID is `ID_UNSET` (and `tag` is blank).


spacecowboy:
thank you @svenjacobs !

Issue #443, opened 2024-12-05
Older articles becoming "Found nothing to fetch"


asl97:
### Checklist

- [X] I have used the search function for [**OPEN**](https://github.com/spacecowboy/feeder/issues) issues to see if someone else has already submitted the same bug report.
- [X] I have **also** used the search function for [**CLOSED**](https://github.com/spacecowboy/feeder/issues?q=is%3Aissue+is%3Aclosed) issues to see if the problem is already solved and just waiting to be released.
- [X] I will describe the problem with as much detail as possible.
- [X] If the bug only to occurs with a certain feed, I will include the URL of that feed.

### App version

2.7.0

### Where did you get the app from

F-Droid

### Android version

13 TKQ1.221114.001 (1.0.2.0.TKHMIXM) (POCO Official/Xiaomi HyperOS)

### Device model

POCO F3

### First occurred

Long ago (sorry)

### Steps to reproduce

1. open older feed item
2. see it says "Found nothing to fetch"

### Expected behaviour

It shows the article that is cached,

by changing the settings to show 10 lines, it suggest there is a cached version.

![Screenshot_2024-12-05-11-49-49-386_com nononsenseapps feeder](https://github.com/user-attachments/assets/cddc7003-a2c9-4f09-bc42-8a29f15defc8)

### Current behaviour

It shows "Found nothing to fetch"

![Screenshot_2024-12-05-11-49-51-773_com nononsenseapps feeder](https://github.com/user-attachments/assets/c76f97f2-11b2-443e-b7d3-b0bdf1390311)

asl97:
To clarify, I can press the article button to fetch the full article with no issue.

It's the "summary"/article that is suppose to be in the feed itself for offline reading that isn't loading for older feed item and showing 'Found nothing to fetch'.

https://github.com/user-attachments/assets/6e1562c8-ede8-4486-bad4-63eda8ed983e

spacecowboy:
You have enabled "fetch full by default" and that is why you are getting an error. feeder can't find the full article to fetch anymore.

when you press the button you switch to the "non-full" version which is always available.

do not enable fetch full by default for feeds which do not need it (like feeder news)

asl97:
My settings are off
![Screenshot_2024-12-06-15-05-05-292_com nononsenseapps feeder](https://github.com/user-attachments/assets/fd6282c9-4d3f-4c49-91ec-5753be2ce283)

[...]

PR #447, opened 2024-12-07
[Hotfix] Show snippet of feed instead of "Found nothing to fetch"

asl97:
For feed if for some reason cache was cleared.

Ideally, at least the non full article shouldn't be written into the cachedir as it is a key part of the whole "Offline reading" mentioned in Features.

https://developer.android.com/training/data-storage/app-specific#internal-remove-cache
> Caution: When the device is low on internal storage space, Android may delete these cache files to recover space. So check for the existence of your cache files before reading them.

I did accidentally put a huge git repo recently into a syncthing folder which I deleted afterwards and that might had cause the article to get deleted

This hotfix is not ideal but at least it's something, as I doubt it would be possible to retrieve the lost data. screenshot:
Screenshot_20241208-011326

I would appreciate if you could fix the tag on https://github.com/spacecowboy/Feeder/issues/443 too
Having an situation where the feed get lost ('Found nothing to fetch') is a bug.

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/Repository.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/Repository.kt
@@ -881,6 +881,7 @@ data class Article(
     val link: String? = item?.link
     val feedDisplayTitle: String = item?.feedDisplayTitle ?: ""
     val title: String = item?.plainTitle ?: ""
+    val snippet: String = item?.plainSnippet ?: ""
     val enclosure: Enclosure =
         item?.enclosureLink?.let { link ->
             Enclosure(
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ArticleViewModel.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ArticleViewModel.kt
@@ -206,8 +206,8 @@ class ArticleViewModel(
                             }
                         } else {
                             Log.e(LOG_TAG, "No default file to parse")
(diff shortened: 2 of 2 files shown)


spacecowboy (changes requested):
So the problem here is that you are removing the feedback to the user that something went wrong.

The current situation is certainly not the best, but this will lead to confusing bug reports where sometimes the snippet is shown instead of the real text.

spacecowboy:
Instead, could you perhaps update where the articles are saved?

https://github.com/spacecowboy/Feeder/blob/d69df4caef9a860390870a22ab59e640477f4faa/app/src/main/java/com/nononsenseapps/feeder/util/FilePathProvider.kt#L37

asl97:
> Instead, could you perhaps update where the articles are saved?
> 
> https://github.com/spacecowboy/Feeder/blob/d69df4caef9a860390870a22ab59e640477f4faa/app/src/main/java/com/nononsenseapps/feeder/util/FilePathProvider.kt#L37

This is just a quick hot fix, and changing the location wouldn't help for already deleted saved article.

Not to mention migrating exisiting articles from the cachedir to another location is a bit out of my league, where the maintainer of this project with more knowledge would be more suited to do.

asl97:
> So the problem here is that you are removing the feedback to the user that something went wrong.
> 
> The current situation is certainly not the best, but this will lead to confusing bug reports where sometimes the snippet is shown instead of the real text.

IMO, `Found nothing to fetch` is not much of a feedback, but if you want a more clear feedback for the user, sure.

Screenshot_20241211-213922

spacecowboy:
@asl97 just run `./gradlew ktlintformat` to fix the ktlint stuff

spacecowboy (approved):

PR #490, opened 2025-01-10
Build APK for specific architectures

yukibtc:
In addition to the universal APK, build specific APK for `arm64-v8a` and `armeabi-v7a` architectures. This allow to download a smaller APK if the device use one of these arch. `x86_64` is used by emulators while `x86` is pretty rare to find (<1% of devices uses this arch) so I excluded them.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -107,6 +107,15 @@ android {
 //                signingConfig = signingConfigs.getByName("shareddebug")
 //            }
         }
+
+        splits {
+            abi {
+                isEnable = true
+                reset()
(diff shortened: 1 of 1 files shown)


dluvian:
Android Studio will lint this when "x86_64" is not included:
`To properly support ChromeOS, your Android application should have an x86 and/ or x86_64 binary as part of the build configuration. To fix the issue, ensure your files are properly optimized for ARM; the binary translator will then ensure compatibility with x86. Alternatively, add an abiSplit for x86 within your build. gradle file and create the required x86 dependencies.`

Issue #464, opened 2024-12-19
List Feature


dayvydrn:
### Checklist

- [X] I have used the search function for [**OPEN**](https://github.com/spacecowboy/feeder/issues) issues to see if someone else has already submitted the same feature request.
- [X] I have **also** used the search function for [**CLOSED**](https://github.com/spacecowboy/feeder/issues?q=is%3Aissue+is%3Aclosed) issues to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.
- [X] I will describe the problem with as much detail as possible.
- [X] This request contains only one single feature, **not** a list of multiple (related) features.

### App version

2.7.3

### Where did you get the app from

Google Play

### Feature or improvement you want

Remove advertising information, i.e. when activated or read aloud, do not mention advertising information, codes or links

Option to differentiate by cores between news titles, i.e., cores changed in the list of selected feeds so as not to confuse the reader

Read titles (headlines) through read aloud

Channel of offers on sale

Automatic jump between different feed sources when not selecting all feeds

Reduce to smaller font and more define the main titles

Social network style buttons to favorite by bookmark, share, etc

dayvydrn:
Enable the option to download images when reading the article

Translation of articles in other languages

Search in the top bar to find articles

check-cyber:
Translation and Search are very useful desired features.
Translate option should be in the menu so that it will translate the titles of all displayed articles in the feed to another language. On opening the article, it should also be translated to same language.
![Uploading Picsart_25-01-06_23-22-19-565.jpg…]()

dayvydaraujo:
Disable a specific feed temporarily

Inplement Proxy VPN for data traffic block case

PR #647, opened 2025-02-28
feat: added a search function

### LINKED ISSUE #368: Global or title search function
### Checklist

### App version

2.6.31

F-Droid

If you have time, I hope you can add a global search function for all subscriptions, or a basic title search function.

### Why should this be added?

When I had a lot of subscriptions, I had to spend a lot of time looking for the old subscriptions I needed.

_No response_

Friendly-Banana:
Allows searching in titles and snippets, one of the features mentioned in #464.

I am new to Kotlin and Compose, so please let me know if there's a better way to do things.

fixes #368

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
@@ -143,6 +143,12 @@ class FeedItemStore(
         }
 
         when {
+            filter.search.isNotEmpty() -> {
+                append("AND (\n")
+                append("plain_title LIKE ?\n").also { args.add("%${filter.search}%") }
+                append("OR plain_snippet LIKE ?\n").also { args.add("%${filter.search}%") }
+                append(")\n")
+            }
             onlySavedArticles -> append("AND bookmarked = 1\n")
             feedId > ID_UNSET -> append("AND feed_id IS ?\n").also { args.add(feedId) }
             tag.isNotEmpty() -> append("AND tag IS ?\n").also { args.add(tag) }
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
@@ -56,6 +56,7 @@ class FeedItemStore(
         tag: String,
         minReadTime: Instant,
         filter: FeedListFilter,
+        search: String,
     ): Flow<Int> {
         val queryString = StringBuilder()
         val args = mutableListOf<Any?>()
@@ -65,7 +66,7 @@ class FeedItemStore(
             append("LEFT JOIN feeds ON feed_items.feed_id = feeds.id\n")
             append("WHERE\n")
 
-            rawQueryFilter(filter, args, minReadTime, feedId, tag)
+            rawQueryFilter(filter, search, args, minReadTime, feedId, tag)
         }
 
         return dao.getPreviewsCount(SimpleSQLiteQuery(queryString.toString(), args.toTypedArray()))
@@ -77,6 +78,7 @@ class FeedItemStore(
         minReadTime: Instant,
         newestFirst: Boolean,
         filter: FeedListFilter,
+        search: String,
     ): Flow<PagingData<FeedListItem>> =
         Pager(
             config =
@@ -95,7 +97,7 @@ class FeedItemStore(
                 append("LEFT JOIN feeds ON feed_items.feed_id = feeds.id\n")
                 append("WHERE\n")
 
-                rawQueryFilter(filter, args, minReadTime, feedId, tag)
+                rawQueryFilter(filter, search, args, minReadTime, feedId, tag)
 
                 when (newestFirst) {
                     true -> append("ORDER BY $FEED_ITEM_LIST_SORT_ORDER_DESC\n")
@@ -112,6 +114,7 @@ class FeedItemStore(
 
     private fun StringBuilder.rawQueryFilter(
         filter: FeedListFilter,
+        search: String,
         args: MutableList<Any?>,
         minReadTime: Instant,
         feedId: Long,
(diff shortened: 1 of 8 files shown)


app/src/main/res/values-en/strings.xml
```diff
@@ -2,4 +2,5 @@
 <resources>
 
     <string name="by_author_on_date">by %1$s, %2$s</string>
+  <string name="search_noun">Search</string>
```
spacecowboy, line 5: Don't add strings to english locale. English is the default locale.

Also pretty sure that 'search' is a verb in this context

app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
```diff
@@ -138,6 +138,7 @@ class SettingsStore(
                 saved = sp.getBoolean(PREFS_FILTER_SAVED, false),
                 recentlyRead = sp.getBoolean(PREFS_FILTER_RECENTLY_READ, true),
                 read = sp.getBoolean(PREFS_FILTER_READ, true),
+                search = "",
```
spacecowboy, line 141: This is not the right place for this. The search query is not a setting.

Should probably be a mutableState in Repository.kt, considering that it is needed later when passed to FeedItemStore

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
```diff
@@ -495,145 +509,393 @@ fun FeedScreen(
...
+                                                contentDescription = closeMenuText
+                                                role = Role.Button
+                                            },
+                                )
+                            }
+                        }
+                    }
+                }
+
+                PlainTooltipBox(tooltip = { Text(stringResource(R.string.search_noun)) }) {
```
spacecowboy, line 704: IMO the search button should definitely be the first button in the toolbar so move that up

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
```diff
@@ -495,145 +509,393 @@ fun FeedScreen(
...
+            if (viewState.searchBarVisible) {
+                SearchBar(
+                    expanded = true,
+                    onExpandedChange = { onShowSearchBar(it) },
+                    inputField = {
+                        SearchBarDefaults.InputField(
+                            query = viewState.filter.search,
+                            onQueryChange = { filterCallback.searchFor(it) },
+                            onSearch = {
+                                onShowSearchBar(false)
```
spacecowboy, line 521: I'm not sure I think it's a good idea to close the search bar and leave the filter in place.

A user might make a search, press "enter", and then have no idea why all articles aren't showing much later.
```suggestion
```

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
```diff
@@ -495,145 +509,393 @@ fun FeedScreen(
...
+                                filterCallback.searchFor(it)
+                            },
+                            expanded = true,
+                            onExpandedChange = { onShowSearchBar(it) },
+                            placeholder = { Text("query") },
+                            leadingIcon = { Icon(Icons.Default.Search, contentDescription = "search") },
```
spacecowboy, line 527: contentDescriptions need translated string as well.

It should also describe the image. In this case "search icon" would be more appropriate, or maybe even "null" (there is no point in including the icon in a screen reader because it is purely decorative)

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
```diff
@@ -495,145 +509,393 @@ fun FeedScreen(
...
+                            trailingIcon = {
+                                IconButton(
+                                    onClick = {
+                                        onShowSearchBar(false)
+                                        filterCallback.searchFor("")
+                                    },
+                                ) {
+                                    Icon(Icons.Default.Close, contentDescription = "cancel search")
```
spacecowboy, line 535: Needs translated string for contentDescription.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
```diff
@@ -495,145 +509,393 @@ fun FeedScreen(
...
                         IconButton(
-                            onClick = { onShowFilterMenu(true) },
+                            onClick = { onShowSearchBar(true) },
                         ) {
                             Icon(
-                                Icons.Default.FilterList,
-                                contentDescription = stringResource(R.string.filter_noun),
+                                Icons.Default.Search,
+                                contentDescription = stringResource(R.string.search_noun),
+                                tint = if (viewState.filter.search.isEmpty()) androidx.compose.material3.LocalContentColor.current else MaterialTheme.colorScheme.primary,
```
spacecowboy, line 712: Subtle to change the icon color when a search filter is applied. Probably too subtle... I'm leaning towards ensuring the search bar can't be closed when non-empty (or without being cleared)

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
```diff
@@ -495,145 +509,393 @@ fun FeedScreen(
...
+                        SearchBarDefaults.InputField(
+                            query = viewState.filter.search,
+                            onQueryChange = { filterCallback.searchFor(it) },
+                            onSearch = {
+                                onShowSearchBar(false)
+                                filterCallback.searchFor(it)
+                            },
+                            expanded = true,
+                            onExpandedChange = { onShowSearchBar(it) },
+                            placeholder = { Text("query") },
```
spacecowboy, line 526: "query" should be a translated string. I think it should just be "Search" too

app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
```diff
@@ -143,6 +143,12 @@ class FeedItemStore(
         }
 
         when {
+            filter.search.isNotEmpty() -> {
```
spacecowboy, line 146: I think instead of baking it into the filter, a new parameter should be added to the function

app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedItemStore.kt
```diff
@@ -143,6 +143,12 @@ class FeedItemStore(
...
+                append("AND (\n")
+                append("plain_title LIKE ?\n").also { args.add("%${filter.search}%") }
```
spacecowboy, line 148: So this will be problematic if the user searches for something with a percent sign. Like `50%`.

Something like this should be done:

```
val sanitizedSearch = filter.search
    // User probably means a literal backslash
    .replace("\\", "\\\\")
    // User probably means a literal percent sign
    .replace("%", "\\%")
    // User probably means a literal underscore
    .replace("_", "\\_")
```

And then also specify escape for the like, e.g.

```
append("plain_title LIKE ? ESCAPE '\\'\n").also { args.add("%$sanitizedSearch%") }
```

Same for next LIKE as well.

I realize this is an issue for the tag as well, but I think it's less of a problem there (but feel free to santize that too!)

spacecowboy (changes requested):
Thanks for doing this! I've left some feedback.

It doesn't look great on Tablets but I think the value it brings is important enough not to let that block it:

tablet_search

spacecowboy (approved):

spacecowboy:
Thanks for you work on this! I think many will be happy to see it.

PR #656, opened 2025-03-07
fix: made feed ordering case insensitive

Friendly-Banana:
Properly sorts the feeds in the sidebar, ignoring case.
Not applied to tags for now, because these are user-defined. 
Is this a feature you'd consider including?

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/db/room/FeedDao.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/db/room/FeedDao.kt
@@ -63,7 +63,7 @@ interface FeedDao {
 
     @Query(
         """
-        SELECT 
+        SELECT
             id,
             COALESCE(NULLIF(custom_title, ''), title) as title,
             notify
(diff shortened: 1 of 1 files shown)


spacecowboy:
Thanks for the PR!

This is a good change, but I think you should use Sqlite's collate feature. example:

    ORDER BY display_name COLLATE NOCASE

then there is no need to include a new column in the queries.

Friendly-Banana:
I tried that, but it fails with `e: [ksp] /home/runner/work/Feeder/Feeder/app/src/main/java/com/nononsenseapps/feeder/db/room/FeedDao.kt:160: There is a problem with the query: [SQLITE_ERROR] SQL error or missing database (no such column: feed_id)`

spacecowboy:
Hmm indeed. This seems to be a quirk of Sqlite. This looks like a solution (wrap it in a select)

```
            -- wrap in select so we can collate in order
            select * from (
                -- all items
                select $ID_ALL_FEEDS as id, '' as display_title, '' as tag, '' as image_url, sum(unread) as unread_count, 0 as expanded, 0 as sort_section, 0 as sort_tag_or_feed
                from feeds_with_items_for_nav_drawer
                -- starred
                union
                select $ID_SAVED_ARTICLES as id, '' as display_title, '' as tag, '' as image_url, sum(bookmarked) as unread_count, 0 as expanded, 1 as sort_section, 0 as sort_tag_or_feed
                from feeds_with_items_for_nav_drawer
                where bookmarked
                -- tags
                union
                select $ID_UNSET as id, tag as display_title, tag, '' as image_url, sum(unread) as unread_count, tag in (:expandedTags) as expanded, 2 as sort_section, 0 as sort_tag_or_feed
                from feeds_with_items_for_nav_drawer
                where tag is not ''
                group by tag
                -- feeds
                union
                select feed_id as id, display_title, tag, image_url, sum(unread) as unread_count, 0 as expanded, case when tag is '' then 3 else 2 end as sort_section, 1 as sort_tag_or_feed
                from feeds_with_items_for_nav_drawer
                where tag is '' or tag in (:expandedTags)
                group by feed_id
            ) as combined
            -- sort them
            order by sort_section, tag, sort_tag_or_feed, display_title collate nocase
```

PR #658, opened 2025-03-08
lang: updated Turkish Translation

mikropsoft:
I re-translated the file by syncing it with the English string file, so it might look complicated. While there are 2 spaces on the left side of the English strings in the translations, the existing Turkish file has 4 spaces. I left it as it should be. Since I’m not using Weblate, I’m opening a PR; if you don’t want to merge it, I’d appreciate it if you could at least upload it to Weblate.


spacecowboy:
I don't mind PRs for translations but you've broken the build. The app doesn't compile with your changes.

I believe the reason is there are now apostrophes in the strings. For example:

```
<string name="add_feed_search_hint">Besleme URL'si</string>
```

These must be escaped. Here is an example of that from the french strings:

```
<string name="failed_to_fetch_full_article_missing_body">Il n\'y a rien à récupérer</string>
```

mikropsoft:
Hmm, I thought I had done these, but I forgot. I was aware of this requirement. I'll change it tomorrow.

By the way, the edit permissions on the PR should be open, so you can fix it if you want.

mikropsoft:
@spacecowboy done.

spacecowboy:
> @spacecowboy done.

Thanks!

PR #714, opened 2025-04-20
fix: added ability to set timeout in OpenAI config

### LINKED ISSUE #682: Error when summarizing if server took too long to process
### Checklist

### App version

2.10.1

F-Droid

Yes

### Android version

15 (crDroid 11)

### Device model

Redmi Note 10

### First occurred

couple weeks ago?

### Steps to reproduce

1. Set up llama-server (if you want to reproduce, run it on a slow machine, or run it on CPU with 1 thread only)
2. Go to settings > API Key, set:
- API Key (if you set one)
- URL (eg. http://127.0.0.1:8080/v1)
- Model ID (if URL correctly typed it should normally show one model that you loaded in llama-server)
3. Open up one of the articles, click ••• and click Summarize

### Expected behaviour

It should wait patiently

### Current behaviour

After pressing Summarize (no matter if parse full article was pressed or not), if after 30 seconds llama-cpp isn't finished, Feeder will throw error like you can see below

Image

There's no place to write additional things, so i'll write below this instead...

llama-cpp is one of the self-hostable/local LLM solution, i'm running it locally on my phone through Termux, i got ~3 tokens per second with Llama 3.2 1B Q3 model which is wayy slower than what normally acceptable, but if i dont have to give my data to other parties or requiring internet connectivity its worth little bit of slow speed.

Imo maybe there should be a way to disable or adjust the socket timeout in the settings, maybe also add cancel button in case it does took too long, also if possible add the real-time text typing instead of having to wait until it finished then show all the summarized text at once, this would be useful for self hosted local LLM that runs slower

### Logs

_No response_

KevinSJ:
I had quite some trouble deciding when to coerce the input values. And decided to do coerce when the input field loses focus.

@spacecowboy Please let me know if you are okay with this implementation.

Will close #682

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
@@ -724,6 +724,7 @@ enum class SwipeAsRead(
 data class OpenAISettings(
     val modelId: String = "",
     val baseUrl: String = "",
+    val timeout: String = "30",
     val azureApiVersion: String = "",
     val azureDeploymentId: String = "",
     val key: String = "",
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
@@ -15,7 +15,6 @@ import com.nononsenseapps.feeder.ui.compose.settings.FontSelection
 import com.nononsenseapps.feeder.ui.compose.settings.getFontSelectionFromPath
 import com.nononsenseapps.feeder.util.FilePathProvider
 import com.nononsenseapps.feeder.util.PREF_MAX_ITEM_COUNT_PER_FEED
-import com.nononsenseapps.feeder.util.filePathProvider
 import com.nononsenseapps.feeder.util.getStringNonNull
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
@@ -463,6 +462,7 @@ class SettingsStore(
                 baseUrl = sp.getStringNonNull(PREF_OPENAI_URL, ""),
                 azureApiVersion = sp.getStringNonNull(PREF_OPENAI_AZURE_VERSION, ""),
                 azureDeploymentId = sp.getStringNonNull(PREF_OPENAI_AZURE_DEPLOYMENT_ID, ""),
+                timeoutSeconds = sp.getInt(PREF_OPENAI_REQUEST_TIMEOUT_SECONDS, 30),
             ),
         )
     val openAiSettings = _openAiSettings.asStateFlow()
@@ -476,6 +476,7 @@ class SettingsStore(
             .putString(PREF_OPENAI_URL, value.baseUrl)
             .putString(PREF_OPENAI_AZURE_VERSION, value.azureApiVersion)
             .putString(PREF_OPENAI_AZURE_DEPLOYMENT_ID, value.azureDeploymentId)
+            .putInt(PREF_OPENAI_REQUEST_TIMEOUT_SECONDS, value.timeoutSeconds.coerceIn(30, 600))
             .apply()
     }
 
@@ -604,6 +605,7 @@ const val PREF_OPENAI_MODEL_ID = "pref_openai_model_id"
 const val PREF_OPENAI_URL = "pref_openai_url"
 const val PREF_OPENAI_AZURE_VERSION = "pref_openai_azure_version"
 const val PREF_OPENAI_AZURE_DEPLOYMENT_ID = "pref_openai_azure_deployment_id"
+const val PREF_OPENAI_REQUEST_TIMEOUT_SECONDS = "pref_openai_request_timeout_seconds"
 
 /**
  * Appearance settings
@@ -724,7 +726,7 @@ enum class SwipeAsRead(
 data class OpenAISettings(
     val modelId: String = "",
     val baseUrl: String = "",
-    val timeout: String = "30",
+    val timeoutSeconds: Int = 30,
     val azureApiVersion: String = "",
     val azureDeploymentId: String = "",
     val key: String = "",
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
@@ -476,7 +476,7 @@ class SettingsStore(
             .putString(PREF_OPENAI_URL, value.baseUrl)
             .putString(PREF_OPENAI_AZURE_VERSION, value.azureApiVersion)
             .putString(PREF_OPENAI_AZURE_DEPLOYMENT_ID, value.azureDeploymentId)
-            .putInt(PREF_OPENAI_REQUEST_TIMEOUT_SECONDS, value.timeoutSeconds.coerceIn(30, 600))
+            .putInt(PREF_OPENAI_REQUEST_TIMEOUT_SECONDS, value.timeoutSeconds)
             .apply()
     }
 
--- a/app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
@@ -42,7 +42,7 @@ class OpenAIClientDefault(
 private fun OpenAISettings.toOpenAIConfig(): OpenAIConfig =
     OpenAIConfig(
         token = key,
-        timeout = Timeout(socket = timeout.toInt().seconds),
+        timeout = Timeout(socket = timeoutSeconds.seconds),
         logging = LoggingConfig(logLevel = LogLevel.Headers, sanitize = !BuildConfig.DEBUG),
         host = toOpenAIHost(withAzureDeploymentId = false),
         httpClientConfig = {
--- a/app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
@@ -42,7 +42,7 @@ class OpenAIClientDefault(
 private fun OpenAISettings.toOpenAIConfig(): OpenAIConfig =
     OpenAIConfig(
         token = key,
-        timeout = Timeout(socket = timeoutSeconds.seconds),
+        timeout = Timeout(socket = timeoutSeconds.coerceIn(30, 600).seconds),
         logging = LoggingConfig(logLevel = LogLevel.Headers, sanitize = !BuildConfig.DEBUG),
         host = toOpenAIHost(withAzureDeploymentId = false),
         httpClientConfig = {
--- a/app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
@@ -3,6 +3,7 @@ package com.nononsenseapps.feeder.openai
 import com.aallam.openai.api.chat.ChatCompletion
 import com.aallam.openai.api.chat.ChatCompletionRequest
 import com.aallam.openai.api.core.RequestOptions
+import com.aallam.openai.api.http.Timeout
 import com.aallam.openai.api.logging.LogLevel
 import com.aallam.openai.api.model.Model
 import com.aallam.openai.client.LoggingConfig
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
```diff
@@ -724,6 +724,7 @@ enum class SwipeAsRead(
 data class OpenAISettings(
     val modelId: String = "",
     val baseUrl: String = "",
+    val timeout: String = "30",
```
spacecowboy, line 727:
```suggestion
    val timeoutSeconds: Int = 30,
```

app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
```diff
@@ -40,6 +42,7 @@ class OpenAIClientDefault(
 private fun OpenAISettings.toOpenAIConfig(): OpenAIConfig =
     OpenAIConfig(
         token = key,
+        timeout = Timeout(socket = timeout.toInt().seconds),
```
spacecowboy, line 45:
```suggestion
        timeout = Timeout(socket = timeoutSeconds.coerceIn(30..600).seconds),
```

app/src/main/java/com/nononsenseapps/feeder/ui/compose/settings/OpenAISection.kt
```diff
@@ -219,6 +228,34 @@ fun OpenAISectionEdit(
...
             },
         )
+
+        TextField(
+            modifier =
+                Modifier.fillMaxWidth().onFocusChanged {
+                    if (!(it.isFocused || isTimeoutInputValid)) {
+                        timeoutString = timeoutString.toInt().coerceIn(30, 600).toString()
+                    }
+                },
```
spacecowboy, line 238:
```suggestion
                Modifier.fillMaxWidth(),
```

Wrong place for this. The onValueChange would have been better, but I see no reason to do any of that in the UI component since you have proper error checking here. But on the backend tough..

app/src/main/java/com/nononsenseapps/feeder/ui/compose/settings/OpenAISection.kt
```diff
@@ -219,6 +228,34 @@ fun OpenAISectionEdit(
...
+            keyboardOptions = KeyboardOptions.Default.copy(keyboardType = KeyboardType.Number),
+            supportingText = {
+                if (!isTimeoutInputValid) {
+                    Text(stringResource(R.string.time_out_validation_error))
+                }
+            },
+            onValueChange = { input ->
+                timeoutString = input.filter { it.isDigit() }
+                if (isTimeoutInputValid) {
+                    onEvent(OpenAISettingsEvent.UpdateSettings(current.copy(timeout = timeoutString)))
```
spacecowboy, line 253:
```suggestion
                    onEvent(OpenAISettingsEvent.UpdateSettings(current.copy(timeout = timeoutString.toInt())))
```

spacecowboy (changes requested):
Hey thanks for doing this. I've added some suggestions.

Nice work with the error checking and error labeling.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/settings/OpenAISection.kt
```diff
@@ -150,6 +154,11 @@ fun OpenAISectionEdit(
     }
 
     var modelsMenuExpanded by remember { mutableStateOf(false) }
+    var timeoutString by remember { mutableStateOf(current.timeout) }
+    val isTimeoutInputValid =
+        timeoutString.isNotEmpty() &&
+            timeoutString.isDigitsOnly() &&
```
spacecowboy, line 160: Maybe add a `trim()` call and not only `isdigitsonly`. a pet peeve of mine is when apps and sites refuses to remove spaces from input fields.

spacecowboy:
Another thing,  you're never actually storing the setting value, so it will be wiped when you restart the app.

You need to add it to SettingsStore.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/settings/OpenAISection.kt
KevinSJ: I've moved the coerceIn call to settingsStore. The change did make the value that's storing into settingsStore correct.

However, when the user opens the openAI settings again, they will see the invalid value they had typed (and the error message), this will be confusing to the user but it is somewhat expected, cause the timeoutString was never updated to reflect the clamped value.

I also attempted the coerceIn call on timeoutString within onValueChange, but that will interfere with user input -- when they type 6, the input field will be concerce immdiately to 30, which is also not ideal. Would you have any other suggestions on this?

app/src/main/java/com/nononsenseapps/feeder/ui/compose/settings/OpenAISection.kt
```diff
@@ -219,6 +225,30 @@ fun OpenAISectionEdit(
...
+            supportingText = {
+                if (!isTimeoutInputValid) {
+                    Text(stringResource(R.string.time_out_validation_error))
+                }
+            },
+            onValueChange = { input ->
+                timeoutString = input.filter { it.isDigit() && !it.isWhitespace() }
+                if (timeoutString.isNotEmpty()) {
+                    onEvent(OpenAISettingsEvent.UpdateSettings(current.copy(timeoutSeconds = timeoutString.toInt())))
+                }
```
spacecowboy, line 247:
```suggestion
                if (isTimeoutInputValid) {
                    onEvent(OpenAISettingsEvent.UpdateSettings(current.copy(timeoutSeconds = timeoutString.toInt())))
                    timeoutString = "${timeoutString.toInt()}"
                }
```

Best to use the same logic in here as is used for the validation. This also ensures the `timeoutString` is in sync with the value which is persisted.

app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
```diff
@@ -476,6 +476,7 @@ class SettingsStore(
             .putString(PREF_OPENAI_URL, value.baseUrl)
             .putString(PREF_OPENAI_AZURE_VERSION, value.azureApiVersion)
             .putString(PREF_OPENAI_AZURE_DEPLOYMENT_ID, value.azureDeploymentId)
+            .putInt(PREF_OPENAI_REQUEST_TIMEOUT_SECONDS, value.timeoutSeconds.coerceIn(30, 600))
```
spacecowboy, line 479:
```suggestion
            .putInt(PREF_OPENAI_REQUEST_TIMEOUT_SECONDS, value.timeoutSeconds)
```

An incorrect value should never be allowed to be put into the datastructure in the first place. And this coercion doesn't fix possible issues present in the `OpenAISettings` object

app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
```diff
@@ -40,6 +42,7 @@ class OpenAIClientDefault(
 private fun OpenAISettings.toOpenAIConfig(): OpenAIConfig =
     OpenAIConfig(
         token = key,
+        timeout = Timeout(socket = timeoutSeconds.seconds),
```
spacecowboy, line 45:
```suggestion
        timeout = Timeout(socket = timeoutSeconds.seconds).coerceIn(30, 600),
```

I think this is the best place to put the coercion since it's where it is actually used

app/src/main/java/com/nononsenseapps/feeder/ui/compose/settings/OpenAISection.kt
spacecowboy: > However, when the user opens the openAI settings again, they will see the invalid value they had typed (and the error message), this will be confusing to the user but it is somewhat expected, cause the timeoutString was never updated to reflect the clamped value.

I've added a comment which should address this. The value is never changed if it is not valid. Closing settings and coming back should load the stored value.

app/src/main/java/com/nononsenseapps/feeder/openai/OpenAIClient.kt
KevinSJ: I think this is where the openai settings for feeder is converted to OpenAI client library config? So I thought putting the cocercion here will not save the correct value after cocercion into settingsStore, that's why I put the cocercion to settingsStore. I can try again later today.

spacecowboy: This will not "save" the coercion but that is not the point.

The point is to ensure a reasonable value is used for the connection even if there is "garbage" in the settings file

spacecowboy: You validation rules will ensure values are in the proper range but bugs (now and especially in the future) are possible

app/src/main/java/com/nononsenseapps/feeder/archmodel/SettingsStore.kt
KevinSJ: I do agree with you but I can't find any other places where I can call coerceIn and fix the `OpenAISettings` object properly beforehand.

If calling it as part of onValueChange, it will be called on every keystroke, and that will interfere user input, and you also don't like calling it as part of onFocusChange which I did initially.

spacecowboy: Just remove the coercion from here and it's fine

app/src/main/java/com/nononsenseapps/feeder/ui/compose/settings/OpenAISection.kt
spacecowboy: Oops. I seem to have accidentally marked this as resolved. Unresolved it.

spacecowboy:
Just fix the last two comments and I think this is good to be merged :+1:

app/src/main/java/com/nononsenseapps/feeder/ui/compose/settings/OpenAISection.kt
KevinSJ: But isn't the input not used **at all** in this case and will be ignored all the time? When you have time, can you please checkout the branch and try in a simulator please? I don't think we need to rush this..

image


This is how the app works on my phone after accepting the change:

https://github.com/user-attachments/assets/35b03b5a-209e-4b79-96d4-423a90eee2b5

spacecowboy: Ah. You're right. There are two issues.

- `isTimeoutInputValid` is a constant. It's value never changes. It needs to be a remembered value using `timeoutString` as its key.
- `onValueChange` should always update the timeoutstring and the `onEvent` should be removed.

the `onEvent` call should only happen when the user presses the save button.

I'll check out the code later tonight and have a deeper look.

spacecowboy:
@KevinSJ I made a PR to your branch (since you do not allow me to push to your branch directly). Please have a test and see how you feel that works

https://github.com/KevinSJ/Feeder/pull/1

KevinSJ:
@spacecowboy Thanks a lot for your help on this. 

I did run into an edge case -- enter 600 first and then delete it to 6, and then save it. When you open the settings again, it's showing 60. I think from the code perspective, this makes sense, cause onValueChange is called on each keystroke, so it's always going to save and restore to the last valid value in the textfield. But from a user perspective, this behavior can feel a bit arbitrary (personallly, I was expecting to see 600).

I think this can be fixed by moving the onEvent to the save button as you mentioned before (it does seem a bit complex)? I can take a look tomorrow or we can merge this as-is if you okay with the behavior.

spacecowboy:
> @spacecowboy Thanks a lot for your help on this.
> 
> I did run into an edge case -- enter 600 first and then delete it to 6, and then save it. When you open the settings again, it's showing 60. I think from the code perspective, this makes sense, cause onValueChange is called on each keystroke, so it's always going to save and restore to the last valid value in the textfield. But from a user perspective, this behavior can feel a bit arbitrary (personallly, I was expecting to see 600).
> 
> I think this can be fixed by moving the onEvent to the save button as you mentioned before (it does seem a bit complex)? I can take a look tomorrow or we can merge this as-is if you okay with the behavior.

Yes it is an edge case, but we have the same problem with the entire UI here. So I think we can merge as is if you are OK with it.

Rewriting the entire onEvent logic for this screen seems a bit out of scope for this PR

KevinSJ:
> > @spacecowboy Thanks a lot for your help on this.
> > I did run into an edge case -- enter 600 first and then delete it to 6, and then save it. When you open the settings again, it's showing 60. I think from the code perspective, this makes sense, cause onValueChange is called on each keystroke, so it's always going to save and restore to the last valid value in the textfield. But from a user perspective, this behavior can feel a bit arbitrary (personallly, I was expecting to see 600).
> > I think this can be fixed by moving the onEvent to the save button as you mentioned before (it does seem a bit complex)? I can take a look tomorrow or we can merge this as-is if you okay with the behavior.
> 
> Yes it is an edge case, but we have the same problem with the entire UI here. So I think we can merge as is if you are OK with it.
> 
> Rewriting the entire onEvent logic for this screen seems a bit out of scope for this PR

That sounds good to me, I'll merge once you approve, thank you!

PR #735, opened 2025-05-06
allow searching in feeds

gkuznik:


spacecowboy:
Just want to note that I've seen the PR. I just need to think if this is wanted or not

PR #796, opened 2025-06-04
feat: Android 10 (sdk29) is now the oldest supported version

spacecowboy:
Removing Conscrypt means relying on platform HTTPS support and TLS1.3 was enabled in SDK29 (Android 10). Hence bumping to that.

This was released in 2019. So in effect supporting devices which received their last OS update 6 years ago.

In Google play, June 2nd 2025, 1.82% of users are on a version of Android older than 10.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -34,10 +34,11 @@ android {
     defaultConfig {
         applicationId = "com.nononsenseapps.feeder"
         // The version fields are set with actual values to support F-Droid
-        // In Play variant, they are overriden and taken from git.
+        // In Play variant, they are overridden and taken from git.
         versionCode = 3691
         versionName = "2.11.5"
-        minSdk = 23
+        // TLS1.3 is enabled in Android 10 (29) and above
+        minSdk = 29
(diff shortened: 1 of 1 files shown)


gsurrel:
Hello, and thank you for Feeder!

Unfortunately, I am part of the 1.82% (or even more, as I do not have the Play Store installed, so I may not be counted). How big is Conscrypt? Is it problematic to keep it? (I do not want to suggest to have two parallel release branches, because of the maintenance burden)

Of course, it's your app, so you have the final say in what you choose to support.

spacecowboy:
> Hello, and thank you for Feeder!
> 
> Unfortunately, I am part of the 1.82% (or even more, as I do not have the Play Store installed, so I may not be counted). How big is Conscrypt? Is it problematic to keep it? (I do not want to suggest to have two parallel release branches, because of the maintenance burden)
> 
> Of course, it's your app, so you have the final say in what you choose to support.

Size is not the issue. The problem is that if forces you to use specific combinations of libraries or it will just crash the entire app.

You can keep using the previous version. That won't stop working.

Issue #772, opened 2025-05-27
Move bug report button to settings


ghost:
### Checklist

- [x] I have used the search function for [**OPEN**](https://github.com/spacecowboy/feeder/issues) issues to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for [**CLOSED**](https://github.com/spacecowboy/feeder/issues?q=is%3Aissue+is%3Aclosed) issues to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.
- [x] I will describe the problem with as much detail as possible.
- [x] This request contains only one single feature, **not** a list of multiple (related) features.

### App version

v2.11.5

### Where did you get the app from

Google Play

### Feature or improvement you want

The bug report button should be moved to settings.

### Why should this be added?

I think settings is a better place for it.

spacecowboy:
Fixed in #892

PR #892, opened 2025-08-19
feat: moved bug report button to settings

crazicus:
This PR is intended to address issue #772. 

**Task:**
Move the Submit bug report button from the overflow menu in the Feed Screen's Top Bar to the Settings screen, as this is a more logical place for this action to exist.

**Solution:**
Removed the code for the menu item in FeedScreen.kt, any handling code, and trailing imports. 
Created a new `ActionSetting` composable for the Bug Report button
- this is essentially an `ExternalSetting`, only without the `currentValue` passed as the subtitle. It might make more sense to just modify this Composable to accept an optional param for the currentValue
Cleaned up warnings in the affected files

**Testing:**
I ran this manually on an emulator to confirm behavior. Screenshots display the new location of the bug report button.
Screenshot_20250818_235250 Screenshot_20250818_235350 
[Screen_recording_20250818_235415.webm](https://github.com/user-attachments/assets/011fc656-f93f-4582-a4ad-2defcec10fdd)


spacecowboy:
LGTM

PR #908, opened 2025-09-03
fix: localized numbers in the drawer

ebraminio:
This is a localization change similar to this change https://github.com/AntennaPod/AntennaPod/pull/4032 which won't have any effect in English but improves the other language that have their own digits.

Before:
image

After:
image


spacecowboy:
Good change!

But you should not use the default Locale in Android. A user is able to change the locale dynamically.

Instead you should do something like this (in Java):

```
getResources().getConfiguration().getLocales().get(0)
```

See for example: https://stackoverflow.com/questions/14389349/android-get-current-locale-not-default

ebraminio:
@spacecowboy Applied, retrieved it from LocalConfiguration so that compose can react to languages changes instantly also.

spacecowboy:
thanks!

PR #929, opened 2025-09-15
feat: improved swipe to dismiss experience

Coopra:
I use Feeder every day and wanted to contribute by improving the swipe to dismiss experience. The changes are as follows:

- Migrated away from using the deprecated `swipeable` modifier to the purpose built M3 control `SwipeToDismissBox`.
- Added lazy layout item animations for improved UX.
- Added haptic feedback so the user knows when they've reached the swipe threshold.
- Tried to reduce accidental swipes (a personal pain point) by disabling swipe gestures when a scroll is detected. This is a common issue in Jetpack Compose but will hopefully be addressed by Google soon through [their work on gesture disambiguation](https://issuetracker.google.com/issues/252334353).

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCompact.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCompact.kt
@@ -62,13 +62,10 @@ fun FeedItemCompact(
     modifier: Modifier = Modifier,
     imageWidth: Dp = 64.dp,
 ) {
-    Surface {
+    Surface(modifier = modifier.height(IntrinsicSize.Min)) {
         Row(
             horizontalArrangement = Arrangement.spacedBy(4.dp),
-            modifier =
-                modifier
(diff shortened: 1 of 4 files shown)


spacecowboy:
Thanks!

I installed it in the emulator a noticed something

* Make sure filter is displaying `recently read`
* Drag your finger slowly to the side

1. When you reach the end, the color of the background switches to the wrong one
2. When you release, it just stays swiped to the side. It does not toggle state or snap back

Screenshot:

swipebug

spacecowboy:
Sorry it took me a while to get time to look at this. This seems good to me now. Thanks for doing it!

One nit-pick is that the vibration ticks when the item snaps back again (when showing read items) but that is not a blocker

Coopra:
> Sorry it took me a while to get time to look at this. This seems good to me now. Thanks for doing it!
> 
> One nit-pick is that the vibration ticks when the item snaps back again (when showing read items) but that is not a blocker

Thanks! I'll look into fixing that.

PR #937, opened 2025-09-21
fix: don't provide haptic feedback when swipe position snaps back

Coopra:
Don't perform haptic feedback if the swipe position automatically snapped back (and thus crossed the threshold again) as a result of the user swiping to read/unread an article when the filter is set to 'Read' or 'Recently read'.


spacecowboy (approved):

spacecowboy:
Thanks @Coopra !

PR #940, opened 2025-09-22
fix: now able to swipe open the nav drawer again

### LINKED ISSUE #939: No longer possible to open drawer by swiping right
### Steps to reproduce

* Use `Only from end` ans the setting for swipe to mark as read

* Try to swipe left to right in the feed list

### Expected behaviour

The navigation drawer should be opened

### Current behaviour

Nothing

Coopra:
This PR restores the ability to open the nav drawer with a swipe gesture again. This was achieved by reverting to the old approach of having an invisible `Box` sit above the content and handle swipe detection for feed items, which could be independently sized down to allow for drag interception for the nav drawer, without impacting the visual size of the feed items.

We still move away from the deprecated `swipeable` modifier in favor of `anchoredDraggable`, and keep the UX improvements for swiping to dismiss feed items that were made previously.

fixes #939

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCompact.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCompact.kt
@@ -20,7 +20,7 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalDensity
-import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewLightDark
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import coil3.compose.AsyncImage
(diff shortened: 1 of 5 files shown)


spacecowboy:
Tested it and found a slight bug

* if you only swipe the item a little bit, the bg alpha and icon are in "off" mode, but the item toggles state when you release anyway

I do agree that the delay in introduced does feed less satisfying.

Coopra:
>Tested it and found a slight bug
>
> if you only swipe the item a little bit, the bg alpha and icon are in "off" mode, but the item toggles state when you release anyway

@spacecowboy I haven't been able to reproduce this bug, let me know if it's still occurring for you.

I'm happy with how things feel now, so I am going to publish.

spacecowboy:
> > Tested it and found a slight bug
> > if you only swipe the item a little bit, the bg alpha and icon are in "off" mode, but the item toggles state when you release anyway
> 
> @spacecowboy I haven't been able to reproduce this bug, let me know if it's still occurring for you.
> 
> I'm happy with how things feel now, so I am going to publish.

Thanks. Sorry for the delay. Will take a look as soon as I can

spacecowboy:
But at the moment there does not seem to be any differences compared to master?

Coopra:
master should still have the issue where you can't swipe open the navigation drawer, whereas these changes enable that again.

spacecowboy:
must have been a github issue. now I can see the changes. .

spacecowboy:
Attaching a video of some bugs. If you only swipe a little, the item is still marked as read even though threshold hasn't been reached


https://github.com/user-attachments/assets/1e075237-8840-4d9d-9d94-93974d5167e9

Coopra:
Thanks, the video was super helpful! The issue was caused by how I was setting up the anchors when `swipeAsRead` was set to `SwipeAsRead.ONLY_FROM_END`; it should be fixed now!

spacecowboy:
Tested and it seems good to me now. Thanks for your help on this @Coopra

PR #978, opened 2025-11-11
Adding `ru` metadata translation

Korb:
Russian text is currently region-independent. ru_RU ≡ ru_BY ≡ ru_KZ, etc.
The `ru` folder covers all Russian-language Android locales via fallback.
`ru-RU` only works for `ru_RU`; others see `en-US`.

Official F-Droid example — https://f-droid.org/docs/All_About_Descriptions_Graphics_and_Screenshots/#fastlane-structure


Korb:
I was wrong.

The F-Droid client displays the description from the `ru-RU` folder for all Russian-language locales (ru_*), including, for example, `ru-BY`, if the `ru` folder is missing. This is due to the specific way metadata is processed in fdroidserver and the client: the server stores localized data under the `ru-RU` key in index-v2.json, and the client uses it as the base for the `ru` language, ignoring regional differences or applying Android's extended fallback (since Android 7+), where an available regional variant (ru-RU) is considered suitable for other regions of the same language in the absence of an exact match or the base `ru`.

Thus, `ru-RU` is a universal variant, suitable for both Google Play and F-Droid, which correctly applies it to all Russian-language locales in the absence of "ru".

And vice versa, at the moment this is even a harmful change, because you still need to have the `ru-RU` requirement for Google Play, which will entail the presence of two files with identical content without any benefit. There will always be a risk that the translation will be included in one of them and not included, or included with a different wording, in the other.

PR #985, opened 2025-11-15
fix: clarified that the app supports many AI integrations and not only OpenAI

StellarStoic:
Changing title **OpenAI integration** to **AI integration** and adding additional info in the AI integration settings. This will let users know they can add other AI providers, not just OpenAI. 

image

image


spacecowboy:
Thank you @StellarStoic ! Could you also run `./gradlew ktlintformat` to fix the lint errors?

spacecowboy:
Thanks again @StellarStoic

PR #996, opened 2025-12-02
feat: change opacity of read items

### LINKED ISSUE #379: Add clarity when a article is read
### Checklist

### App version

2.6.32

Other

When an article has been read, the entire card's color should dim instead of the title being unbolded. 

### Why should this be added?

Currently it is very hard to see which articles have been read at a glance, since the difference between a bold and non-bold title is slim. This makes it much easier to tell, especially for users with accessibility needs. 

Once a user reads an article, the entire card on the feeds page should be dimmed in color.

Jiogo18:
Set the opacity of the card at 75% when the article is read.
fix #379 
The value can be discussed. I think 50% is too much, 75% may be not enough.

The alpha value can be defined somewhere else. However, I don't know if/how to move this modifier to Typography.
https://m3.material.io/styles/color/roles#63d6db08-59e2-4341-ac33-9509eefd9b4f Perhaps Color.kt?

I also experimented with a more generic approach by changing the alpha of the text in Typography.kt, but this doesn't change the opacity of the website logo. https://github.com/Jiogo18/Feeder/commit/afa8b57a599d3cdbc08a992ea0bd9e5cac2c41f5

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCard.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedItemCard.kt
@@ -109,12 +109,6 @@ fun FeedItemCard(
                                     val pxWidth = maxWidth.roundToPx()
                                     Size(pxWidth, (pxWidth * 9) / 16)
                                 }
-                            val alpha =
-                                if (item.unread) {
-                                    1f
-                                } else {
-                                    0.74f
-                                }
                             AsyncImage(
                                 model =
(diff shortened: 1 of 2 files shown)


Jiogo18:
Result at 75%
Screenshot_20251202-010947_FeederD

I don't expect this PR to be accepted as is, the integration can be greatly improved.

spacecowboy:
Also apologies for the delay, I've been without computer

Jiogo18:
Yes, it changes the opacity of the entire card. Removing it.

PR #1006, opened 2025-12-17
fix: fixed favicon lookup for feeds linking to their own RSS URL

CryYxx:
Some feeds use <link> pointing to the RSS endpoint instead of the site root.
When fetching site metadata, Feeder currently uses the full URL, which prevents favicon discovery.
This change derives the base site URL (scheme + host) before fetching metadata.


spacecowboy:
Thank you @CryYxx

PR #1009, opened 2025-12-21
fix: invisible SearchBar

### LINKED ISSUE #975: Search Bar goes black
### Checklist

### App version

2.16.0

Google Play

No

### Android version

15

### Device model

Samsung xcover 6 pro

### First occurred

Today

### Steps to reproduce

Select Search Icon
Search bar goes full Grey & does not accept input. 

### Expected behaviour

Accept Search input

### Current behaviour

_No response_

### Logs

_No response_

luk-dushaj:
I looked into issue #975 and traced the problem to how the SearchBar expansion state was being handled.

Both the SearchBar and its inputField were configured with expanded = true. This appears to cause the SearchBar to start in an already-expanded state, which results in the search UI being present but not visibly rendered.

Setting SearchBar(expanded = false) while leaving expansion controlled by the inputField restores the expected behavior: the SearchBar is visible in its collapsed state and expands correctly on interaction.

This change only adjusts the initial expansion state and does not alter any search logic.

fixes #975

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
@@ -499,7 +499,7 @@ fun FeedScreen(
         toolbarActions = {
             if (viewState.searchBarVisible) {
                 SearchBar(
-                    expanded = true,
+                    expanded = false,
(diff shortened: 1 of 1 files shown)


spacecowboy:
Thank you @luk-dushaj for fixing this!

Sorry for the delay

PR #1018, opened 2026-01-07
feat: added widget

### LINKED ISSUE #384: Newsfeed Widget
### Checklist

### App version

2.6.33

F-Droid

Large customizable widget to emulate similar experience as with Google News Feed

### Why should this be added?

For people who got used to Google News Feed Swipe on start page as me, it would finally make me uninstall Google News Feed and only use Feeder as my main News source. Have seen lots of requests on reddit from people searching for a proper Google News Feed alternative. Feeder was mentioned a lot but it unfortunately has no widget support so people did not find a solution with Feeder.

UI should be similar to Google Newsfeed. Small Banners you can scroll through. When clicking on an article, the feedly app or a Browser Tab should open with full details of the article / post.

MatthewTighe:
This PR adds a newsfeed widget using Jetpack Glance for #384. It:

- establishes a separate "current feed and tag" for the widget
- allows customization of the display feed and tag through a settings activity
- allows refreshing the widget through a "refresh button"

There are a couple areas that could potentially use more polish, but I wanted to get this up to at least start a conversation as to whether it is worth waiting for them.

Areas of potential improvement:
1. the widget preview never renders, just loads forever
2. lack of tests - Glance is allegedly unit testable, but because this access `LocalContext`, it would need to provided by something like Robolectric, which I did not think was worth importing for that alone. there's otherwise very little logic here, but happy to eventually add tests for the Repository etc if preferred
3. I can't get the "Refreshing" text to actually display for some reason - though it does if I make that the only content. I'll probably keep messing with this at some point
4. widget isn't resizable

LMK your thoughts on whether any of these are worth baking longer - happy to keep trucking on this but it's obviously been slow going. Assume we will just squash these commits as well?

[Screen_recording_20260106_222913.webm](https://github.com/user-attachments/assets/0802ecd2-49d4-4ad4-9039-08685bbe85f8)

Fixes #384

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/Repository.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/Repository.kt
@@ -397,8 +397,7 @@ class Repository(
         combine(
             currentWidgetFeedAndTag,
             feedListFilter,
-            search,
-        ) { feedAndTag, feedListFilter, search ->
+        ) { feedAndTag, feedListFilter ->
             val (feedId, tag) = feedAndTag
             FeedListArgs(
                 feedId = feedId,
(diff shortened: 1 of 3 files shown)


MatthewTighe:
@spacecowboy thanks so much for taking a look! I've applied your patch feedback. 

For the memory crash, I've tried a couple things:
1. setting a target size to the coil ImageRequest. I doubt this did much, but I'm wondering if perhaps oversized bitmaps were being loaded and then just visually scaled down.
2. I changed the format used for loading the image to the lowest density one.

That said, I haven't run into the crash in my local testing. Could you either see if you can still reproduce, or perhaps pass along a repro OPML export?

I've kept the patch applying your feedback separate for now, assuming we will squash and merge eventually

spacecowboy:
> @spacecowboy thanks so much for taking a look! I've applied your patch feedback.
> 
> For the memory crash, I've tried a couple things:
> 
> 1. setting a target size to the coil ImageRequest. I doubt this did much, but I'm wondering if perhaps oversized bitmaps were being loaded and then just visually scaled down.
> 2. I changed the format used for loading the image to the lowest density one.
>
> That said, I haven't run into the crash in my local testing. Could you either see if you can still reproduce, or perhaps pass along a repro OPML export?

The feed I tried with was Feeder's default feed for the app's changelog which automatically gets added on fresh installs: https://news.nononsenseapps.com/

> I've kept the patch applying your feedback separate for now, assuming we will squash and merge eventually

Yes it will get squashed on merge. You forgot to remove the `search` for `getWidgetFeedListItems` though


I will try the new version soon

MatthewTighe:
Whoops, looks like `getWidgetFeedListItems` had become unused anyway so I removed it. I'm still not able to reproduce the crash you're seeing unfortunately

https://github.com/user-attachments/assets/2deb8fe7-7e8c-49d0-9b51-e6e918938d52

spacecowboy:
I'll take another look. 

@MatthewTighe do you feel it is ready now from your perspective and testing?

spacecowboy:
It seems to work nicely now, and the resizing works smoothly :+1: 

One thing I noted is that the margins are un-even. There is a margin on the left but not on the right. But this is not a blocker by any means. This can be fixed later.

MatthewTighe:
Ha, I am not the most skilled when it comes to UX 😅 I often have to rely on the professionals for that piece of sign-off. Happy to take a look if you'd like. If it's not a blocker, is there anything else you need from me before this is merged?

spacecowboy:
Then I think we should just merge. Not letting perfect be the enemy of good.

Thanks for the contribution @MatthewTighe !

MatthewTighe:
:raised_hands: awesome, thanks for helping me get it over the finish line! looking forward to using it in the wild :grin:

PR #1036, opened 2026-02-08
feat: when YouTube feed is added, suggest additional channel specific feeds

xn-7492:
If any YouTube channel rss feed is found in the feed search, additional feeds will be suggested for: Videos only, Shorts only and Livestreams only. More info on YouTube playlist prefixes [here](https://stackoverflow.com/a/79656556)

feed-search


spacecowboy (approved):
Thank you for the PR! Looks good to me

PR #1065, opened 2026-03-13
fix: added divider after 'All feeds' and 'Saved articles' in the NavDrawer. 

JaredTweed:
This is much more intuitive, especially for new users.


spacecowboy:
Would you mind providing a screenshot of your change?

JaredTweed:
Here you go!

Screenshot_20260316-092325
Screenshot_20260316-092343

spacecowboy:
Seems ok to me

Issue #1063, opened 2026-03-13
Widget isnt showing any content after adding certain feeds


efb4f5ff-1298-471a-8973-3d47447115dc:
### Checklist

- [x] I have used the search function for [**OPEN**](https://github.com/spacecowboy/feeder/issues) issues to see if someone else has already submitted the same bug report.
- [x] I have **also** used the search function for [**CLOSED**](https://github.com/spacecowboy/feeder/issues?q=is%3Aissue+is%3Aclosed) issues to see if the problem is already solved and just waiting to be released.
- [x] I will describe the problem with as much detail as possible.
- [x] If the bug only to occurs with a certain feed, I will include the URL of that feed.

### App version

2.18.0

### Where did you get the app from

F-Droid

### Are you using the "Parse full article" feature?

No

### Android version

GrapheneOS Android 16

### Steps to reproduce

1. Add https://www.privacyguides.org/news/rss/ as feed and name it Privacy Guides Privacy & Security News
2. Add https://www.privacyguides.org/rss/ as feed and name it Privacy Guides Articles
3. Go to widget and press on the refresh feed button

Sometimes the app it self gets a bit unresponsive or crashes

### Expected behaviour

To see articles

### Current behaviour

![Image](https://github.com/user-attachments/assets/2dddc74b-d2e6-4645-944c-a66b9d4c2af5)

spacecowboy:
@MatthewTighe is this something you could look into?

spacecowboy:
@efb4f5ff-1298-471a-8973-3d47447115dc if possible, please provide adb logs. You can filter on `FEEDER`

MatthewTighe:
Yes, I've been tagged in another issue which I want to look at. I've noticed some other behaviors that I want to improve which I thought I'd accounted for as well. I'll try to spend some time with it this weekend. Thanks for the ping!

efb4f5ff-1298-471a-8973-3d47447115dc:
@MatthewTighe you can look at all my issues because they are all related to the widget :)

efb4f5ff-1298-471a-8973-3d47447115dc:
seems to be resolved

PR #1066, opened 2026-03-15
fix: fixed widget not showing any content

MatthewTighe:
This fixes a number of issues, I would suggest inspecting the changes in commit order.

Probably the most substantial change is to how data is loaded into the widget for fixing #1063. Previously, we were trying to re-use the paging data behavior from the main app, but this isn't really supported by Glance. I removed the DataStore that was intended to hack around this limitation and instead introduced some methods through the Repo stack to load a deterministic subset of feed data (limited to 50 items).

The UX changes include:
1. adding a background to cards to try and differentiate between items
2. aligning the title with the icon
3. allowing text wrapping of the title and description
4. applying a color tint to the read items (though this was the later patch)

Now that the previews are fixed, I can easily show some of these changes 😁 

image

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/db/room/FeedItemDao.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/db/room/FeedItemDao.kt
@@ -162,6 +162,57 @@ interface FeedItemDao {
         FROM feed_items
         LEFT JOIN feeds ON feed_items.feed_id = feeds.id
         WHERE feed_id IS :feedId
+          AND block_time IS NULL
+        ORDER BY $FEED_ITEM_LIST_SORT_ORDER_DESC
+        LIMIT 50
+        """,
+    )
+    fun widgetPreviewsByFeed(feedId: Long): Flow<List<PreviewItem>>
+
+    @Query(
+        """
+        SELECT $PREVIEW_COLUMNS
+        FROM feed_items
+        LEFT JOIN feeds ON feed_items.feed_id = feeds.id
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nononsenseapps/feeder/db/room/FeedItemDao.kt
```diff
@@ -156,6 +156,57 @@ interface FeedItemDao {
...
+        SELECT $PREVIEW_COLUMNS
+        FROM feed_items
+        LEFT JOIN feeds ON feed_items.feed_id = feeds.id
+        WHERE feed_id IS :feedId
+          AND block_time IS NULL
+        ORDER BY $FEED_ITEM_LIST_SORT_ORDER_DESC
+        LIMIT 50
+        """,
+    )
+    fun widgetPreviewsByFeed(feedId: Long): Flow<List<PreviewItem>>
```
spacecowboy: Very very reasonable 👍

spacecowboy (approved):

spacecowboy:
Thank you @MatthewTighe !

omer-ash33:
The widget is still not working for me. It's stuck on loading. I'm using Kvaesitso launcher, but I also tried on other launchers including the default one, the widget works in none of them. I have the latest version of Feeder installed from F-droid.

leo1mml:
Same here, the widget is blank and loading forever.
I'm on oneUI from Samsung. I tested the Playstore version and the Fdroid one.

Issue #809, opened 2025-06-13
option to manually open a single article in default browser instead of custom tab


ilf:
### Checklist

- [x] I have used the search function for [**OPEN**](https://github.com/spacecowboy/feeder/issues) issues to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for [**CLOSED**](https://github.com/spacecowboy/feeder/issues?q=is%3Aissue+is%3Aclosed) issues to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.
- [x] I will describe the problem with as much detail as possible.
- [x] This request contains only one single feature, **not** a list of multiple (related) features.

### App version

2.11.5

### Where did you get the app from

F-Droid

### Feature or improvement you want

Currently, items opened "in web view" (article view, top bar, second icon from the right) are opened in "custom tab".

I would like to open these items in "default browser".

Settings -> Reader has the option "open links with" with possible values "custom tab" and "default browser". But this only works for links in items text, not the items themselves.

### Why should this be added?

I believe we can and should make opening links in browsers behave consistently.

In my setup, "custom tab" doesn't work for me.

### Screenshots / Drawings / Technical details

This can probably be achieved by either expanding the option "open links with" to cover "in web view", too.

Or we could create a new option for "open links in vew view with".

spacecowboy:
Feeder will never support web views again (which aren't related to your default browser at all).

are you refering to opening articles in default browser by defaulr? which is a setting you can pick for each feed individually, as well as globally.

or are you referring specifically to the button in article view? which currently only opens articles with custom tab (which IS related to your default browser)?

the feature request appears to be conflating several things, making it unclear

ilf:
I mean specifically the button in article view.

When I long-press that, a tooltip pops up which sais "Open in Web view".
I took the name from that, sorry if the actual term is something else.

spacecowboy:
> I mean specifically the button in article view.
> 
> When I long-press that, a tooltip pops up which sais "Open in Web view". I took the name from that, sorry if the actual term is something else.

ah. That text should be updated maybe.

so to summarize for my future self: add a way to manually open a single article in default browser

ilf:
Yes! I changed the title to that :)

ilf:
I tried to look into this again.

(This issue it pretty annoying for me. Instead of one click in Feeder, this requires me to make three clicks in Feeder and additional five in the browser.)

I believe this is the line that launches this:

[...]

PR #1068, opened 2026-03-16
New setting to open article in browser instead of custom tab

xn-7492:
- New setting to open article in browser instead of custom tab
- add 'Apply blocklist to summary' setting to opml import/export


spacecowboy:
Please don't do PRs for more than one change at a time. Split one of your changes out into its own PR

xn-7492:
got it, i've moved the second change to a new pr

spacecowboy (changes requested):
Technically, it seems fine.

But I don't like the phrasing `Open items in reader`.  At this moment I don't have a much better suggestion though.

I feel that there is no way of knowing from the SettingsPage that this refers to the button in the menu bar 🤔

xn-7492:
I have renamed it to `Open articles via button in reader with`, is that better?

ilf:
Hi. I wonder what the status of this is.
Is the next task with @xn-7492 or @spacecowboy?
I would love to see https://github.com/spacecowboy/Feeder/issues/809 fixed.
Thanks, both of you!

PR #1096, opened 2026-04-01
fix: update feed url if selfLink has a new address in the original feed

xn-7492:
Currently the feed xml url doen't get updated when it changes in the fetched feed.
RssLocalSync already prevents updating links that contain auth or query params.


spacecowboy:
thanks @xn-7492

GideonBear:
Hi @spacecowboy, just wanted to let you know that this broke my feed. Huginn used `https://` instead of `http://` as the field url for my locally hosted feed, while Huginn doesn't support HTTPS.
This is completely not Feeder's responsibility, but just letting you know in case you want to consider adding a check for that too; there could be more feeds that do this.

spacecowboy:
good to know @GideonBear . could be reasonable to ensure that the protocol doesn't change in a url update

spacecowboy:
@GideonBear see https://github.com/spacecowboy/Feeder/pull/1228

PR #1097, opened 2026-04-03
feat: added stripping of tracking parameters before opening URLs

### LINKED ISSUE #1044: Remove URL Tracking parameter when sharing links from app
### Checklist

### App version

2.17.0

F-Droid

When sharing an article from within Feeder, it adds a URL Tracking Parameter: ?traffic_source=rss

Can we please get rid of that and keep the app more privacy-friendly?

Thank you

### Why should this be added?

Because it improves the app privacy. 

_No response_

mvanhorn:
## Summary

Strips `utm_*` and `traffic_source` query parameters from article URLs when sharing via the share button. The feed source adds these tracking parameters to article links; removing them before sharing keeps shared URLs cleaner.

## Changes

- New: `app/src/main/java/com/nononsenseapps/feeder/util/UrlCleaner.kt` - `stripTrackingParameters()` function that removes `utm_*` and `traffic_source` params while preserving all other query parameters
- `ArticleScreen.kt` (line 144): Apply parameter stripping to the shared article link
- `FeedScreen.kt` (lines 1353, 1587): Apply parameter stripping to both feed list share paths

## How it works

The function parses the URL with `android.net.Uri`, checks if any tracking parameters exist, and rebuilds the URL without them. If no tracking parameters are present or the URL can't be parsed, it returns the original URL unchanged.

Stripped prefixes: `utm_source`, `utm_medium`, `utm_campaign`, `utm_term`, `utm_content`, `traffic_source`.

## Testing

- URL with `?utm_source=rss&id=123` -> shared as `?id=123`
- URL with `?traffic_source=rss` -> shared with no query string
- URL with no tracking params -> shared unchanged
- Malformed URL -> returned as-is

Closes #1044

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/util/UrlCleaner.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/util/UrlCleaner.kt
@@ -1,0 +1,35 @@
+package com.nononsenseapps.feeder.util
+
+import android.net.Uri
+
+/**
+ * Strips common tracking query parameters from URLs before sharing.
+ *
+ * Removes standard UTM parameters and other well-known tracking tags
+ * while preserving the rest of the URL.
+ */
+fun stripTrackingParameters(url: String): String {
+    val uri =
+        try {
+            Uri.parse(url)
+        } catch (_: Exception) {
+            return url
+        }
+
+    val paramNames = uri.queryParameterNames
+    if (paramNames.isEmpty()) return url
+
+    val trackingPrefixes = listOf("utm_", "traffic_source")
+    val hasTracking = paramNames.any { name -> trackingPrefixes.any { name.startsWith(it) } }
+    if (!hasTracking) return url
+
+    val builder = uri.buildUpon().clearQuery()
+    for (name in paramNames) {
+        if (trackingPrefixes.none { name.startsWith(it) }) {
+            for (value in uri.getQueryParameters(name)) {
+                builder.appendQueryParameter(name, value)
+            }
+        }
+    }
+    return builder.build().toString()
+}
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nononsenseapps/feeder/util/UrlCleaner.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import android.net.Uri
+
+/**
+ * Strips common tracking query parameters from URLs before sharing.
+ *
+ * Removes standard UTM parameters and other well-known tracking tags
+ * while preserving the rest of the URL.
+ */
+fun stripTrackingParameters(url: String): String {
```
spacecowboy: Please add a basic unit test

spacecowboy (changes requested):
Added some comments.

Further, CI is red. Please ensure CI passes

mvanhorn:
Fixed all three items:
- Switched to proper imports in both FeedScreen.kt and ArticleScreen.kt
- Added unit test (UrlCleanerKtTest) with 5 test cases
- Refactored UrlCleaner to use java.net.URI instead of android.net.Uri so it's testable without Robolectric

CI should be green now.

spacecowboy:
Please run `./gradlew ktlintformat`

And please look at the output of the CI jobs and fix them

mvanhorn:
Fixed both CI issues in 560c7982:

1. Smart cast - `viewState.articleLink` is an interface property so Kotlin can't smart-cast it after the null check. Captured it in a local `val` first.
2. ktlint - ran `ktlintFormat`, fixed import ordering in `FeedScreen.kt`.

`ktlintCheck` passes locally. Don't have Android SDK set up so couldn't verify compile locally, but the smart cast fix is straightforward.

spacecowboy (approved):

spacecowboy:
thanks @mvanhorn

Issue #411, opened 2024-11-05
Scroll distance isn't saved


mrn:
### Checklist

- [X] I have used the search function for [**OPEN**](https://github.com/spacecowboy/feeder/issues) issues to see if someone else has already submitted the same bug report.
- [X] I have **also** used the search function for [**CLOSED**](https://github.com/spacecowboy/feeder/issues?q=is%3Aissue+is%3Aclosed) issues to see if the problem is already solved and just waiting to be released.
- [X] I will describe the problem with as much detail as possible.
- [X] If the bug only to occurs with a certain feed, I will include the URL of that feed.

### App version

2.6.33

### Where did you get the app from

F-Droid

### Android version

11

### Device model

OnePlus 6T

### Steps to reproduce

1. Open an article
2. Read through some of it (scroll down)
3. close the article
4. return to the article

### Expected behaviour

Scroll distance is remembered and when the article is opened again, it automatically scrolls to the previously saved distance.

### Current behaviour

Opening an article which was partially read starts from the top.

spacecowboy:
Renamed title to properly indicate that this is a feature request.

Absolutely no clue how it could be implemented

PR #1115, opened 2026-04-13
feat: preserve article scroll position when Android recreates the process

boringrgb:
Feeder on Android seems to lose the scroll position in the currently open article. When reading a long form article in Feeder, if one changes context to another app and Android kills the process the behaviour of Feeder is that it goes back to the feeds index and refreshes the feeds. Which interrupts the reading experience and can be disruptive.

This PR keeps note of where in the article the reader was and preserves that across the process wake cycle.

Preserves scroll position in article, only for curretly opened article, related to feature request #411. In that feature it is suggested that Feeder do that for every article, which would involve storing that state somewhere (probbly in the db). The approach suggested in the PR is only for the currently opened article and needs no database chages/migration.

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ArticleScreen.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ArticleScreen.kt
@@ -60,6 +60,7 @@ import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.role
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.compose.runtime.snapshotFlow
 import com.nononsenseapps.feeder.R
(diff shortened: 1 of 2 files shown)


spacecowboy:
Thanks for the PR. That does seem like a fairly elegant solution.

If I understand it correctly, when an article is opened, then at first the scroll position defaults to 0. Then, as the user scrolls, the view model updates every 500ms.

Later, when the user switches articles, the view model gets destroyed and a new view model gets created for the new article, so the scroll position always starts at 0 for a newly opened article.

So it would probably work if Android destroys the process and it gets recreated later.

So better than the state today. Would still be nice to have it save it in such a way that you can go back to the list and then open the same article again and the scroll position would be kept. But that can be kept as a future improvement.

boringrgb:
The going back to the list view and then back to an article preserving reading posisiton should be possible but I think then it would have to be serialized somehow, possibly to the db, which would require a regression test that I didn't want to do. This seemed a little less invasive a change.

PR #1130, opened 2026-05-09
feat: add per-feed article metadata thumbnail enrichment

### LINKED ISSUE #1125: Add support for og:image in articles
Some webcomic feeds use the `og:image` tag, would be good if Feeder could parse and display it.

- [ ] Need example feed to test against

tokenflood:
Some feeds have missing thumbnails. This PR adds a per-feed toggle to fetch `og:image`/`twitter:image` from the article page during RSS sync and use it when an item has no better thumbnail. It also avoids duplicate header images in the article reader.

**OG thumbnail enrichment**

New per-feed toggle in feed settings (default OFF): *Enhance thumbnails from article metadata*. When enabled, each synced item that only has a body-extracted image (or no image at all) gets an extra request to the article URL to look for `og:image`/`twitter:image` meta tags. Only the `<head>` section of the page is streamed and the connection is closed as soon as the `</head>` boundary is hit, so no full page body is downloaded.

Feed-provided thumbnails (media/enclosure) are never replaced, and enrichment only applies to items whose displayed thumbnail is missing or body-derived. If a feed provides a thumbnail that is too small to use and the item falls back to a body image, enrichment is still attempted. Once an item has been upgraded to a non-body thumbnail, it won't be fetched again on later syncs. Relative image URLs are resolved against the final redirected page URL and blank metadata values are ignored.

The toggle survives encrypted feed sync and OPML import/export. There is no separate background job; enrichment happens inline during sync.

**Full-text duplicate image fix**

The article reader previously always showed the header thumbnail for non-body images, which could duplicate the same image inline. It now compares the header image URL against the rendered article content and only hides it when that URL already appears inline, so articles that do not duplicate the hero image in their body still keep their header image.

DB version bumped to 39 with a new `feeds.fetch_og_images` column (default 0). Tests cover thumbnail policy, feed item update merging, reader image dedupe, OG image extraction, head-content streaming, and serialization/export coverage for the new field.

Fixes #1125

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ReaderView.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ReaderView.kt
@@ -92,7 +92,8 @@ fun ReaderView(
     feedTitle: String,
     authorDate: String?,
     image: ThumbnailImage?,
-    isFeedText: Boolean,
+    showHeaderImage: Boolean,
+    contentImageUrls: Set<String>,
     modifier: Modifier = Modifier,
     articleScrollState: ScrollState = rememberScrollState(),
     articleBody: @Composable (indexOffset: Int) -> Unit,
@@ -290,8 +291,7 @@ fun ReaderView(
                 }
             }
 
-            // Don't show image for full text articles since it's typically inside the full article
-            if (isFeedText && image?.fromBody == false) {
+            if (image != null && shouldShowHeaderImage(showHeaderImage, image, contentImageUrls)) {
                 val imageWidth: Int =
                     if (image.width?.let { it < 640 } == true) {
                         // Zero or too small, do not show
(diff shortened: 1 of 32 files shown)


app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ReaderView.kt
```diff
@@ -375,12 +375,19 @@ private fun ReaderPreview() {
...
             ) {}
         }
     }
 }
 
+internal fun shouldShowHeaderImage(
+    showHeaderImage: Boolean,
+    image: ThumbnailImage?,
+    contentImageUrls: Set<String>,
+): Boolean = showHeaderImage && image?.fromBody == false && image.url !in contentImageUrls
```
spacecowboy, line 389: 👍

app/src/main/java/com/nononsenseapps/feeder/util/HtmlUtils.kt
```diff
@@ -33,3 +42,119 @@ fun findFirstImageInHtml(
...
+}
+
+private fun MediaType.isHtml(): Boolean = subtype == "html" || subtype == "xhtml+xml"
+
+/**
+ * Stream only the `<head>` portion of a web page.
+ * Reads just enough bytes to find `</head>`, keeping network usage minimal
+ * (typically 1–2 KB vs 100+ KB for a full page).
+ */
+internal fun streamHeadContent(inputStream: InputStream): String {
```
spacecowboy, line 97: 👍

spacecowboy (commented):
I think it looks good. Thanks @tokenflood 

One question: if the `head` of the article is fetched for the `og:image` and rest of body is not fetched, this does not mess up the okhttp cache later if the full article is fetched? I assume okhttp is smart enough to handle it and you haven't noticed anything weird in your tests I assume?

tokenflood:
> I think it looks good. Thanks @tokenflood
> 
> One question: if the `head` of the article is fetched for the `og:image` and rest of body is not fetched, this does not mess up the okhttp cache later if the full article is fetched? I assume okhttp is smart enough to handle it and you haven't noticed anything weird in your tests I assume?

Good question. This does a normal GET, but the body is closed as soon as `</head>` is found. My understanding is that OkHttp only commits responses to the cache after the body has been fully read, so closing early should not result in a truncated page being stored in the cache. A later full-article fetch should therefore fetch/cache the full page normally.

I’ve also been using this locally for over a month and haven’t noticed any weird cache/full-article behavior.

PR #1131, opened 2026-05-09
feat: upgraded export format so saved articles can be imported again

### LINKED ISSUE #913: Import saved articles.
### Checklist

### App version

2.13.0

Other

Would be really nice to be able to import back my saved articles. Since there's an export. 

I had to reinstall Feeder so saved articles are gone.

### Why should this be added?

I reinstalled Feeder. Was able to export then import my feeds. And Even though there's an option to export saved articles, there's no option to import them back.

_No response_

edd255:
Similarly to the user who created issue #913, I have noticed that Feeder is not able to import saved articles, which is a bit annoying when you switch devices. This PR should fix it. I have build this in Android Studio and tested it in a Google Pixel 9a emulator, and it seems to work.

Fixes #913

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
@@ -182,7 +182,7 @@ fun FeedScreen(
     val di = LocalDI.current
     val savedArticleExporter =
         rememberLauncherForActivityResult(
-            ActivityResultContracts.CreateDocument("text/plain"),
+            ActivityResultContracts.CreateDocument("application/json"),
         ) { uri ->
             if (uri != null) {
                 val applicationCoroutineScope: ApplicationCoroutineScope by di.instance()
@@ -387,7 +387,7 @@ fun FeedScreen(
                     savedArticleExporter.launch(
                         "feeder-saved-articles-${LocalDate.now()}-${
                             LocalTime.now().toSecondOfDay()
-                        }.txt",
+                        }.json",
                     )
                 } catch (_: Exception) {
                     // ActivityNotFoundException in particular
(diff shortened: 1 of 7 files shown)


spacecowboy:
Thanks for the PR @edd255 . I'm afraid this only solves a very small part of the overall feature.

Consider the following:

* User subscribes to a feed which only contains the 10 latest articles.
* User saves one of the articles.
* Time passes, and now the saved article is no longer part of the 10 items in the feed. But it has been marked as saved so Feeder does not autodelete it.
* User exports saved articles
* User reinstalls Feeder and wishes to reimport saved articles

At this point, the acceptance criteria is:

* The saved article which is no longer part of the feed is present in Feeder

And a major obstacle to this is that the existing import does not contain enough information to fully resurrect the item in the database. All the metadata comes from the original feed entry which is no longer available, and Feeder's export is just of the link.

To actually solve this, the exported data would need to be amended and contain the necessary metadata.

edd255:
Thanks for the feedback! :) I will take a look at this in a few days.

This will break backwards-compatibility to some extent, as old exports cannot be imported then. But that seems like a reasonable trade-off.

spacecowboy:
> This will break backwards-compatibility to some extent, as old exports cannot be imported then. But that seems like a reasonable trade-off.

Not much to break. Existing exports of saved articles can't be imported at all. They only exist to allow people to extract their data.

edd255:
I built it and tested it on a real device. It should work now. Importing saved articles from feeds you are not subscribed to also works.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
```diff
@@ -352,12 +364,30 @@ fun FeedScreen(
...
             },
+            onImportSavedArticles = {
+                try {
+                    savedArticleImporter.launch(
+                        arrayOf(
+                            "text/plain",
+                            // This is the mimetype the file may get from older exports
+                            "application/octet-stream",
+                            // But just in case a file isn't named right etc, accept all
+                            "*/*",
```
spacecowboy, line 376: Shouldn't we list `application/json` here for completeness?

spacecowboy (changes requested):
This is great @edd255 . Nice work!

I added a small nit-question inline

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feed/FeedScreen.kt
edd255: Yes, you are right, thank you!

spacecowboy (approved):

spacecowboy:
Thanks again @edd255

Aga-C:
@spacecowboy Can you unblock translations on Weblate?

PR #1132, opened 2026-05-09
feat: enabled arm memory tagging extension

flexxxxer:
makes app exploitation much harder on devices and OS's with MTE support https://source.android.com/docs/security/test/memory-safety/arm-mte#enable-mte-apps

I was using Feeder for around a week and it runs fine for me on GrapheneOS with Memory tagging enabled for an app. Propose to enable it by default.

This change will affect:
- Google Pixel's users with stock OS with Advanced Protection enabled
- Google Pixel's users with GrapheneOS
- Samsung Galaxy users [in some near future](https://www.androidauthority.com/one-ui-9-memory-tagging-extension-spotted-3654992/)
- other devices with other OS's in some not that near future

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -45,6 +45,7 @@
     android:roundIcon="@mipmap/ic_launcher_round"
     android:supportsRtl="true"
     android:theme="@style/AppThemeDayNoActionBar"
+    android:memtagMode="async"
     android:usesCleartextTraffic="true">
 
     <meta-data


spacecowboy:
I don't know. This seems risky:

> We recommend using this mode in production on well tested codebases where the density of memory safety bugs is known to be low, which is achieved by using the SYNC mode during testing. 

And what threat are you mitigating here? Feeder is hardly something worth exploiting.

flexxxxer:
> I don't know. This seems risky:
> 
> > We recommend using this mode in production on well tested codebases where the density of memory safety bugs is known to be low, which is achieved by using the SYNC mode during testing.

If you have no ability to debug an app nor test it on MTE-compatible device nor offload that testing on users (I am) then it is risky. Anyway I didn't achieve any MTE error on GrapheneOS with that app for few Feeder versions.

> And what threat are you mitigating here? Feeder is hardly something worth exploiting.

Well many people do not care about private messages because this "hardly something worth". Same for public media. But for some people spectrum changes, including but not limited with RSS feeds list. Theoretically it is possible to pass maliciously-crafted response for Feeder (Feeder does not have MITM protections - this is expected) and cause in-app memory exploit to, lets say, send all RSS feeds to remote server (well Feeder has access to Feeder's data storage and internet). MTE would make those theoretical (and I hope it would stay that way xD) exploits harder, that's it.

spacecowboy:
Let's just see..

PR #1137, opened 2026-05-18
fix: widget stuck in loading

### LINKED ISSUE #1113: Widget loads indefinitely
### Checklist

### App version

2.18.0

Google Play

Yes

### Android version

16

### Device model

_No response_

### First occurred

After latest update

### Steps to reproduce

1. install widget
2. see indefinite loading

### Expected behaviour

Regardless of how long a feed takes to load, the basics of the widget should still render. The title bar, etc, so that the selected feed can still be changed. 

### Current behaviour

Image

### Logs

_No response_

MatthewTighe:
Okay wow, thanks for your patience while I figured this one out. I finally thought to try a local release build and was able to replicate the issue, which meant I could log what was happening. It would appear that R8 is minifying a work manager no-arg constructor used by Glance under the hood, and that's crashing on newer Android versions. Hopefully this is solved upstream at some point, but this seems to do it for now!

fixes #1113

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -49,6 +49,9 @@
 # Charset detection
 -keep class com.ibm.icu.** { *; }
 
+# For WorkManager - InputMerger subclasses are instantiated by name via reflection
+-keep class * extends androidx.work.InputMerger { <init>(); }
+
 # For Kodein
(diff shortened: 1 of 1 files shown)


spacecowboy:
Thanks a lot @MatthewTighe !

PR #1143, opened 2026-05-18
feat: added on-device local as an option for translation

JaredTweed:
## What does this change?

Added local translation as a provider option. I tested it, and it works. This enables people to translate articles without an API key.

## Checklist

## Screenshots (if UI change)

Screenshot_20260518-173356


spacecowboy:
Looks good. What is bergamot (besides a citrus fruit)?

JaredTweed:
@spacecowboy Sorry, that was the translator I tried to use originally, but I switched it to the on-device version (and I forgot to clean the remnants of that until just now). The PR is now cleaned and there are no unnecessary files.

spacecowboy:
thanks!

Digitalpeer1:
Thanks So much looking forward to using this!

Digitalpeer1:
Just want to preface this by saying I have no software experience please let me know if anything im saying is stupid or to hard to impliment I am thankfull that @JaredTweed has done this!

> @spacecowboy Sorry, that was the translator I tried to use originally, but I switched it to the on-device version (and I forgot to clean the remnants of that until just now). The PR is now cleaned and there are no unnecessary files.

Was so exited to try this I compiled from the Master branch but realised that because Im on Graphine (Im assuming that this uses the Pixel OS proprietary google translate api built into pixels) rather that Bergamot I get the "Local translation is not available on this device" error. found [this](https://support.mozilla.org/en-US/kb/android-translation?as=u) article on how Firefox does this (this is my only refrence for how this is done) and this quote stands out for how Im assuming this might not work on Feeder

"Why can't I access translations on my Android device?

Translation is currently only supported on Android devices that are based on the aarch64 architecture. The reason behind this requirement is that Firefox needs to be able to run WebAssembly to complete the translation.

While we don’t have a comprehensive list of all the Android devices that don’t support translation, in case your device is one of them, you should still see if support is available by visiting WebAssembly and checking if "Fixed Width SIMD" or "Relaxed SIMD" are supported by your Firefox version.

Translation was released in Firefox 129 via a Nimbus experiment, but it runs via the Fixed-Width SIMD status by default with the release of Firefox 130."

Here are my logs if helpful
[FeederD log 75c7e5977adf.txt](https://github.com/user-attachments/files/28075228/FeederD.log.75c7e5977adf.txt)

JaredTweed:
@Digitalpeer1 ah shoot! Sorry about that. I am a big fan of graphene OS, and I will see if I can get that working for you. I originally planned to do it that way as you suggested, but it is definitely a larger project than using the built-in API from Google. I built this using a Pixel 8 with the Pixel OS so I didn't even consider that issue.

PR #1144, opened 2026-05-19
feat: use feed entry id as fallback link.

fictiontoreality:
## What does this change?

Some RSS feeds store entry links in the `<id>` field rather than in a separate `<link>` field. Example: https://jacobin.com/feed

```
...
<entry>
  <id>https://jacobin.com/2026/05/democrats-medicare-for-all-donor-class</id>
...
```

This change takes advantage of this by:
1. if `<link>` field is present, use that as the link.
2. if not, check if the `<id>` field is a valid absolute HTTP/S URL, and if so use that as the link.

## Checklist

- [ ] If the Room schema changed: added a `MIGRATION_N_N+1` in `AppDatabase.kt`
- [ ] If the Room schema changed: added a migration test in `app/src/androidTest/.../db/room/`

## Screenshots (if UI change)
N/A


spacecowboy:
thanks @fictiontoreality . Normally I'd say that we shouldn't cater to such incorrect feeds, but the code is clean and is covered by tests so :+1:

rasidhulalam152-cpu:
Ehierf

PR #1147, opened 2026-05-21
feat: added Bergamot local translation fallback

JaredTweed:
## What does this change?

Adds an app-provided local translation fallback using Bergamot/Mozilla-style models when Android system on-device translation is unavailable.

- Keeps Android `TranslationManager` as the first local backend.
- Adds a Bergamot WASM backend backed by the MPL `@browsermt/bergamot-translator` assets.
- Adds model management for language-pair lookup, English pivoting, SHA-verified downloads, app storage, deletion, and offline reuse.
- Stores app-provided translation models under `filesDir/translation-models/bergamot`.
- Updates Local translation settings copy so unsupported local translation explains the OS path and app-provided model path separately.
- Adds JVM tests for registry parsing, direct pair downloads, pivot pair resolution, checksum failure, status, and deletion.

This is intended to cover devices where Android `TranslationManager` reports unavailable, including GrapheneOS-style environments without Google/Pixel system translation support.

## Verification

- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew ktlintFormat`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew ktlintCheck`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew compileFdroidDebugKotlin`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew testFdroidDebugUnitTest --tests com.nononsenseapps.feeder.localtranslation.BergamotModelManagerTest`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew lintFdroidDebug`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew installFdroidDebug`
- Launched the fdroid debug build on a connected Pixel 8 Pro; no startup crash in logcat.
- Ran a Node smoke test against the vendored Bergamot package, translating `Hallo Welt!` to `Hello world!` with a real downloaded model.

## Checklist

Room schema was not changed.

## Screenshots (if UI change)

No screenshots; UI change is settings copy only.


app/src/main/java/com/nononsenseapps/feeder/localtranslation/TranslationLanguageCodes.kt
```diff
@@ -0,0 +1,32 @@
...
+        "dutch", "nl" -> "nl"
+        "polish", "pl" -> "pl"
+        "russian", "ru" -> "ru"
+        "czech", "cs" -> "cs"
+        "estonian", "et" -> "et"
+        "bulgarian", "bg" -> "bg"
+        "icelandic", "is" -> "is"
+        "norwegian", "nb", "nn" -> "nb"
+        "persian", "fa" -> "fa"
+        "ukrainian", "uk" -> "uk"
```
spacecowboy, line 35: Is this the complete list of supported languages or can more be added over time?

app/src/main/java/com/nononsenseapps/feeder/localtranslation/BergamotWebTranslator.kt
```diff
@@ -0,0 +1,171 @@
...
+    private var initializedRegistryJson: String = ""
+
+    suspend fun translate(
+        content: List<String>,
+        sourceLanguage: String,
+        targetLanguage: String,
+        preserveHtml: Boolean,
+        modelRegistry: List<BergamotModelRegistryEntry>,
+    ): BergamotWebTranslationResult =
+        withContext(Dispatchers.Main.immediate) {
```
spacecowboy, line 55: Why are you running this on the UI thread?

app/src/main/java/com/nononsenseapps/feeder/localtranslation/BergamotWebTranslator.kt
```diff
@@ -0,0 +1,171 @@
...
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicLong
+
+class BergamotWebTranslator(
+    private val application: Application,
```
spacecowboy: Please use kodein dependency injection instead of this constructor argument

app/src/main/java/com/nononsenseapps/feeder/localtranslation/LocalTranslator.kt
```diff
@@ -26,14 +27,18 @@ import org.jsoup.nodes.TextNode
...
 class LocalTranslator(
     override val di: DI,
 ) : DIAware {
     private val application: Application by instance()
+    private val filePathProvider: FilePathProvider by instance()
     private val directExecutor = Executor { runnable -> runnable.run() }
+    private val bergamotModelManager by lazy {
+        BergamotModelManager(filePathProvider.filesDir.resolve("translation-models").resolve("bergamot"))
+    }
+    private val bergamotWebTranslator by lazy { BergamotWebTranslator(application) }
```
spacecowboy: The classes should use dependency injection themselves

```suggestion
    private val bergamotModelManager: BergamotModelManager by instance()
    private val bergamotWebTranslator: BergamotWebTranslator by instance()
```

app/src/main/java/com/nononsenseapps/feeder/localtranslation/BergamotModelManager.kt
```diff
@@ -0,0 +1,317 @@
...
+import okhttp3.Request
+import java.io.File
+import java.security.MessageDigest
+import java.util.Locale
+
+class BergamotModelManager(
+    private val modelRoot: File,
+    private val okHttpClient: OkHttpClient = OkHttpClient(),
+    private val registryUrl: String = DEFAULT_REGISTRY_URL,
+) {
```
spacecowboy: Should use dependency injection here instead for `okHttpClient`.
```suggestion
class BergamotModelManager(
    private val modelRoot: File,
    private val registryUrl: String = DEFAULT_REGISTRY_URL,
): DIAware {
    private val okHttpClient: OkHttpClient by instance()
```

app/src/main/java/com/nononsenseapps/feeder/localtranslation/BergamotWebTranslator.kt
```diff
@@ -0,0 +1,171 @@
...
+            id: Long,
+            message: String,
+        ) {
+            pending.remove(id)?.complete(BergamotWebTranslationResult.Error(message))
+        }
+
+        @JavascriptInterface
+        fun onLog(message: String) {
+            if (BuildConfig.DEBUG) {
+                android.util.Log.d("FeederBergamot", message)
```
spacecowboy:
```suggestion
                logDebug("FEEDER_BERGAMOT", message)
```

Use utility method which strips debugLogs from production releases instead.

Also always put log tags in all caps

spacecowboy (changes requested):
How long does the download take?

I don't think the user gets any progress feedback on the download, right? That would be good to have

JaredTweed:
Addressed the review comments in the latest push.

Changes made:
- Added model-download progress state from `BergamotModelManager` and surfaced it on both the article screen and feed screen.
- Progress is determinate when model sizes are known, and the UI notes first downloads are usually 20-50 MB and may take a few minutes.
- Moved `BergamotModelManager` and `BergamotWebTranslator` to Kodein DI, using the app-provided `OkHttpClient` and `FilePathProvider`.
- Added bounded WebView ready/translation timeouts and batch translation through a JS `translateBatch` API.
- Removed universal file URL access, kept file-to-file access for the trusted asset page to read downloaded model files, and added WebView error logging through `logDebug("FEEDER_BERGAMOT", ...)`.
- Enforced non-blank SHA-256 hashes for model files and replaced temp-file renames with cleanup-safe `Files.move`.
- Documented that `TranslationLanguageCodes` is only aliases, not the full supported model list.
- Expanded model-manager tests for missing registry, unavailable pair, missing hash, status, and deletion behavior.

Verification:
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew ktlintFormat`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew compileFdroidDebugKotlin`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew testFdroidDebugUnitTest --tests com.nononsenseapps.feeder.localtranslation.BergamotModelManagerTest`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew ktlintCheck`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew lintFdroidDebug`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew :app:installFdroidDebug`
- Launched the installed debug app on the connected Pixel 8 Pro; no matching crash lines in recent filtered logcat.

JaredTweed:
Okay, I made a lot of fixes and checked it over and tested it, it should be good now. Each model takes ~3 seconds to download each language for me, and they have loading bars now. I have attached all the new screenshots.

  - BergamotWebTranslator now uses Kodein DI: BergamotWebTranslator(di), with Application by instance().
  - LocalTranslator uses DI for both classes:
    BergamotModelManager by instance() and BergamotWebTranslator by instance().
  - BergamotModelManager uses DI for OkHttpClient: private val okHttpClient: OkHttpClient by instance().
  - Log tag is uppercase: FEEDER_BERGAMOT.

Screenshot_20260521-135736

Screenshot_20260521-135914

Screenshot_20260521-140022

JaredTweed:
@spacecowboy Let me know if you want any changes.

Digitalpeer1:
This works well for me thanks so much for this feature!

JaredTweed:
@Digitalpeer1 Your welcome! Happy I could help, thank you for testing it!

Digitalpeer1:
After a few hours of testing it looks like some articles don't work for some reason. They are all from the same source and the same author I get the "Android system translation unavailable" error again here are the logs if helpful 
[FeederD log fd6cd7c8df21.txt](https://github.com/user-attachments/files/28148586/FeederD.log.fd6cd7c8df21.txt)
and the source and article used: https://kiritgentilal.substack.com/p/a-falsa-modernidade-da-reforma-laboral

app/src/main/java/com/nononsenseapps/feeder/localtranslation/TranslationLanguageCodes.kt
JaredTweed: TranslationLanguageCodes.kt is only a convenience alias map for common names/codes entered in settings. Unknown values are normalized and returned as language so newly added Bergamot registry pairs can work without changing this list.

app/src/main/java/com/nononsenseapps/feeder/localtranslation/BergamotWebTranslator.kt
JaredTweed: Because Android WebView must be created and interacted with on the main/UI thread. The Kotlin side only initializes the WebView and schedules JavaScript calls there. The heavy Bergamot translation work runs inside the WebView’s JavaScript/Web Worker/WASM runtime, not as Kotlin CPU work on the UI thread. The PR now also has WebView init and translation timeouts to avoid indefinite hangs.

JaredTweed:
@Digitalpeer1 do they fail because they do not auto-translate, or they do not translate at all? EDIT: Ah, I see, I have been able to recreate the issue.

JaredTweed:
@Digitalpeer1 Thank you for finding that bug! I fixed it. Let me know if you encounter any others.

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ArticleScreen.kt
```diff
@@ -648,6 +651,58 @@ fun ArticleContent(
...
+                )
+            }
+            Text(
+                text = stringResource(R.string.offline_translation_model_download_hint),
+                modifier = Modifier.padding(top = 8.dp),
+                style = MaterialTheme.typography.bodySmall,
+                color = MaterialTheme.colorScheme.outline,
+            )
+        }
+    }
```
spacecowboy: This progress content seems to be identical to the one defined for FeedScreen.

I think it would be good to extract it as a common `TranslationProgressContent` function. Fine to wrap it in Surface in one place and Card in the other

spacecowboy (changes requested):
One comment on UI reuse. Otherwise looks good.

I'll instal it myself and test it too

app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ArticleScreen.kt
JaredTweed: Good catch, I fixed that just now.

spacecowboy:
OK, I've tried it and I can't get it to work. So I think there are some usability issues to resolve.

For context I'm running a Samsung phone with no Samsung language packs available.

* I'm not sure `Preferred translation language` should be a plain text field. A drop down with valid values would be preferable. If I can't get it to work then I'm not sure regular users would.

* When the field is left blank it seems translation is not available? Why is it possible to enable Local Translation and leave this field blank if it is required?

The toast message was so long it was truncated and pretty useless. I had to add an error log to the View model to see what the problem is:

```
Translation failed: java.lang.IllegalStateException: Android system translation unavailable: 
Local translation does not support en to sv on this device.
                                                                                                    
Android system translation is unavailable, and Feeder does not have an app-provided 
Bergamot model path for en to sv.
                                                                                                    
App model storage: /data/user/0/com.nononsenseapps.feeder.debug/files/translation-models/bergamot
```

I tried a few different language strings: `norwegian`, `danish`, `swedish`, `sv`.

JaredTweed:
You're right! The issue is poor inputs, I will switch it to a drop-down. That will prevent those issues.

spacecowboy:
Thanks @JaredTweed

JaredTweed:
Screenshot_20260525-094559
Screenshot_20260525-094605
Screenshot_20260525-094609
Screenshot_20260525-094702
Screenshot_20260525-101731

JaredTweed:
@spacecowboy Okay, I made those changes, I am ready for you to check them.

spacecowboy:
OK this time

* I toggled `Show only languages with Bergamot models`
* Selected english in the dropdown
* Press translate in a Swedish article

Result was a popup asking me to install `sv->en` in system settings.

This may be by design, but I assumed the feature would download a bergamot model in this case since no Samsung language pack was available.

If it is intentional that Bergamot not be available on Pixel/Samsung phones, then I think we should hide the checkbox (you already identify them). Otherwise someone, like me, will think they can force Bergamot to be used because for reasons they don't want to use Samsung's stuff.

## Trying to use Samsung's translation service

l am looking in settings and I have several language packs installed. In particular both English and Swedish. There are no separate A->B packs. So it seems the Samsung translation integration isn't working quite right because the app keeps prompting me to install `sv->en`

JaredTweed:
Swedish is not an available option for bergamot. I have made it more clear when the option is not available, and must be downloaded via the system settings.

JaredTweed:
I don't have a Samsung, I will see someone who has one in a few weeks, I'll update this then

app/src/main/java/com/nononsenseapps/feeder/localtranslation/LocalTranslator.kt
```diff
@@ -323,9 +544,39 @@ private sealed interface LocalTranslationResult {
...
 
+private fun systemSettingsRequiredMessage(
+    sourceLang: String,
+    targetLang: String,
+): String = "Install $sourceLang -> $targetLang in system settings"
+
+private fun noBergamotModelMessage(
+    sourceLang: String,
+    targetLang: String,
+): String = "No Bergamot model for $sourceLang -> $targetLang. Check system translation languages"
```
spacecowboy: These error messages need to move to String resources. You can see several examples using value interpolation with `%1$s` and `%2$s`

spacecowboy (changes requested):

Digitalpeer1:
Any update on this?

app/src/main/java/com/nononsenseapps/feeder/localtranslation/LocalTranslator.kt
JaredTweed: Fixed in 2b44e301. All user-facing local translation errors now come from string resources, including the interpolated source/target model errors using %1$s and %2$s. I also kept model preparation failures structured by reason and map them to localized text in LocalTranslator.

JaredTweed:
@spacecowboy This is ready for re-review. I merged the current master branch, addressed the latest review request by moving all user-facing local translation errors into string resources (including interpolated language-pair errors), and resolved the remaining addressed review threads. The PR is conflict-free and all GitHub Actions and CircleCI checks, including emulator tests, are passing.

spacecowboy:
Thanks @JaredTweed . I think it's good to merge now

spacecowboy (approved):

spacecowboy:
@JaredTweed an issue has appeared. I may need to revert this PR. It is blocking F-Droid releases: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/47282

JaredTweed:
I confirmed the F-Droid failure is caused by the vendored `bergamot-translator-worker.wasm`: fdroidserver rejects it during the source scan before Gradle runs. I opened #1212 to remove the prebuilt Bergamot runtime and fallback while preserving Android system translation and features merged afterward.

A runtime download would still bypass F-Droid review for executable code, so the safe immediate fix is removal. Bergamot can be restored later once the WASM runtime is reproducibly built from source with an F-Droid-compatible toolchain. The local CI-equivalent checks and `assembleFdroidRelease` pass, and no binary candidates remain under `app/src`.

JaredTweed:
Ignore #1212. We can use #1213 as the fix.

It keeps Bergamot offline translation while removing the prebuilt WASM from the APK: Feeder now asks for explicit consent, clearly warns that the executable download is not built or checked by F-Droid, downloads the pinned 0.4.5 runtime, and verifies its SHA-256 before loading it.

PR #1163, opened 2026-06-10
add setting for max article size

baer842:
add setting for max article size. this fixes an issue, introduced with the oom fix for large articles, that on devices with enough memory that dont crash they can't be displayed anymore


spacecowboy:
Hi thanks for your PR.

First, could you describe the problem you are trying to solve? What feed are you hitting the limit with? It it the memory limit or max element limit?

Second, I don't want to have a setting for this. The heap memory is not that different depending on your RAM. The limit should be such that it works for all devices. The primary limit is the RAM taken by rendering the articles. The download limit has some leeway to adjust depending on what you are seeing.

Third, CI is red because you have failing tests. I won't review anything in detail until the PR is green

baer842:
I use often the full article mode for example with the feed https://www.wired.com/feed/category/security/latest/rss with this nearly every article hits the max mb cap since the new update before it worked perfectly. The main issue for the articles is that i has about 1.3 mb or 1.2 so if its okey for you to raise it from 1 mb to say 2 or 4 it would fix it and the setting would not be needed

spacecowboy:
Yes this is an example of a site which I don't think poses any problems. It's big because it has a huge amount of inline CSS, Javascript, and SVG assets.

I suspect we can safely increase size to 2MB

spacecowboy:
@baer842 please see https://github.com/spacecowboy/Feeder/pull/1164

baer842:
Good to see, will this fix get a new version?

spacecowboy:
Yes, I'll release a new version this weekend

PR #1165, opened 2026-06-12
fix: restored scroll position indicator in the article reader

### LINKED ISSUE #1150: Scroll or progress bar in article reader is missing
### Checklist

### App version

2.20.0

Other

A scroll bar on the right-hand side or a thin progress bar at the top of the article reader. This could also be optional, so enabled/disabled via the in-app settings.

### Why should this be added?

When reading long articles it is nice to know how far in you are and/or how long you have left. A scroll bar or progress bar could solve this by giving an indicator of where you are in the article.

A scroll bar would be preferred as it has the extra utility of being able to quickly go to a different part of the article.

Downloaded directly from GitHub using Obtainium.

mvanhorn:
## What does this change?

Restores the scroll position indicator in the article reader. The `LazyColumn` to `Column` rewrite in #1045 (commit 0416165d, which fixed the selected-text crash) dropped the `my.nanihadesuka` scrollbar wrapper added in 65fef39a, so the reader has had no progress indicator since. The scrolling `Column` is now wrapped in `ColumnScrollbar` bound to the same hoisted `articleScrollState`, with the original thumb styling (`ScrollbarSettings.Default.copy(thumbUnselectedColor = MaterialTheme.colorScheme.outlineVariant)`). The `com.github.nanihadesuka:LazyColumnScrollbar` dependency was still declared and bundled, just unused, so no gradle change is needed.

This does not reintroduce the #1045 crash: the plain `Column` with `verticalScroll` stays, the scrollbar is only an overlay around it.

## Checklist

- [x] Ran `./gradlew ktlintFormat` and committed the result (no formatting changes were produced)
- [ ] If the Room schema changed: added a `MIGRATION_N_N+1` in `AppDatabase.kt` (N/A, no schema change)
- [ ] If the Room schema changed: added a migration test in `app/src/androidTest/.../db/room/` (N/A, no schema change)

## Screenshots (if UI change)

The change adds the same scrollbar overlay the reader had before 0416165d; thumb appears on the right edge of the reader pane while scrolling long articles and is absent for articles that fit on screen.

Fixes #1150

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ReaderView.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/ui/compose/feedarticle/ReaderView.kt
@@ -71,12 +71,12 @@ import com.nononsenseapps.feeder.ui.compose.theme.PreviewTheme
 import com.nononsenseapps.feeder.ui.compose.utils.ProvideScaledText
 import com.nononsenseapps.feeder.ui.compose.utils.ScreenType
 import com.nononsenseapps.feeder.ui.compose.utils.focusableInNonTouchMode
+import my.nanihadesuka.compose.ColumnScrollbar
+import my.nanihadesuka.compose.ScrollbarSettings
 import java.time.format.DateTimeFormatter
 import java.time.format.FormatStyle
 import java.util.Locale
 import kotlin.math.roundToInt
-import my.nanihadesuka.compose.ColumnScrollbar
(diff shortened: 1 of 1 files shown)


spacecowboy:
Thanks @mvanhorn !

mvanhorn:
Thanks @spacecowboy for taking this in. Good to have the article reader scroll position indicator back after the LazyColumn to Column rewrite dropped it.

PR #1173, opened 2026-07-10
feat: added button for renaming tags

### LINKED ISSUE #907: Rename existing label
### Checklist

### App version

2.13.0

Google Play

Add an option, either in settings or directly on a label, to rename the existing tag. This will rename the label, and "move" all feeds on the old label to the new one.

### Why should this be added?

As I add feeds to labels, sometimes I want to update the existing label name to be more inclusive of what I've added. However, you cannot rename the label directly. You can only add each feed to a new label name. This is fine for smaller labels/groups, but it can be a hassle if you have over a dozen feeds for a single label.

Option 1: underneath the three dots, if it's a label and not a single feed (and not "all"), add a button to "Rename Label." When pressed, it will open a dialogue to enter the new name.
Option 2: within "Settings", add a "Manage Labels" section that lists each label. Clicking on each label opens a dialog to rename that specific label.

EmaYYellow:
## What does this change?

This PR implements a feature to rename existing tags to all correlated feeds.

Previously, there was no direct way to modify a tag's name without modifying it in every single feed. This change introduces a contextual "Rename Tag" dialog that updates the tag string across all associated feeds in the local database, ensuring proper state propagation throughout the architecture. The button positioned in the menu only appears when user is viewing a tag, not a feed.

This addresses the requested feature and closes the corresponding issue. (Closes #907)

- Added a custom `RenameTagDialog` Composable component to handle user text input and validation.
- Added a rename button in `FeedScreen` that contextually appears when the user is inside a specific tag view.
- Updated `FeedDao` with a targeted SQL `UPDATE` query to modify the tag column for matching feeds.
- Implemented the `renameTag` execution flow across `FeedStore`, `Repository`, and `FeedViewModel`.
- Added defensive state handling in `Repository`: if the user renames the tag they are currently viewing, the app automatically and seamlessly transitions the active view to the new tag name.
- Added unit tests covering both execution paths (current tag vs. different tag) in `FeedStoreTest` and `RepositoryTest`.
- Did not change the Room database schema (only added a query), so no migration was required.

## Checklist

- [ ] If the Room schema changed: added a `MIGRATION_N_N+1` in `AppDatabase.kt` (N/A - No schema changes)
- [ ] If the Room schema changed: added a migration test in `app/src/androidTest/.../db/room/` (N/A)

### Before
FeederD-BEFORE

### After
FeederD-AFTER

> *Note: Adding a 11th item to the dropdown menu might (like in my case) cause Android to automatically shift the menu position upward on some screen resolutions to prevent it from clipping at the bottom.*

Fixes #907

### DIFF
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedStore.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/FeedStore.kt
@@ -53,6 +53,13 @@ class FeedStore(
 
     suspend fun getDisplayTitle(feedId: Long): String? = feedDao.getFeedTitle(feedId)?.displayTitle
 
+    suspend fun renameTag(
+        oldTag: String,
+        newTag: String,
+    ) {
+        feedDao.renameTag(oldTag, newTag)
+    }
+
     suspend fun deleteFeeds(feedIds: List<Long>) {
         feedDao.deleteFeeds(feedIds)
     }
--- a/app/src/main/java/com/nononsenseapps/feeder/archmodel/Repository.kt
+++ b/app/src/main/java/com/nononsenseapps/feeder/archmodel/Repository.kt
@@ -141,6 +141,17 @@ class Repository(
         settingsStore.setCurrentWidgetFeedAndTag(feedId, tag)
     }
 
+    suspend fun renameTag(
+        oldTag: String,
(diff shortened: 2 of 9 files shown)


spacecowboy:
Thanks! Looks good to me

PR #1182, opened 2026-07-20
fix: honored custom tab preference in notifications

### LINKED ISSUE #879: Clicking notification ignores feed "open with" preference
### Checklist

### App version

2.12.2

Google Play

No

### Android version

Android 16, GrapheneOS

### Device model

Pixel 7a

### First occurred

Always?  At least months ago

### Steps to reproduce

1. Configure a feed to open articles in a custom tab
2. Click on a new article notification for that feed

### Expected behaviour

The article should be opened using a custom tab (or whatever the preference is for that feed)

### Current behaviour

Feeder opens the article using the built-in reader (which is the app default, BTW)

### Logs

_No response_

arpitagarwal1301:
## What does this change?

Makes article-notification taps honor each feed's configured article opener. Custom-tab feeds now open the article in a custom tab, while reader and default-browser behavior remain unchanged.

The custom-tab route continues through the existing notification proxy so the item is marked read and notified and its notification is cancelled. A distinct intent action prevents the content tap from colliding with the notification's separate browser action.

Closes #879

## Checklist

- [x] Ran `./gradlew ktlintFormat`
- [x] Room schema unchanged; no migration required

## Testing

- 9 focused API 30 managed-device notification intent tests
- Full JVM test suite
- F-Droid debug APK assembly
- `ktlintCheck`
- `git diff --check`

## Screenshots

Not applicable; this changes notification routing without changing app UI layout.

### DIFF
--- a/app/src/androidTest/java/com/nononsenseapps/feeder/model/RssNotificationsKtTest.kt
+++ b/app/src/androidTest/java/com/nononsenseapps/feeder/model/RssNotificationsKtTest.kt
@@ -3,8 +3,12 @@ package com.nononsenseapps.feeder.model
 import android.content.Intent
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
+import com.nononsenseapps.feeder.archmodel.ItemOpener
 import com.nononsenseapps.feeder.db.COL_LINK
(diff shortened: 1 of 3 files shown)


spacecowboy:
Thanks

moklese (commented):
Md Moklese

PR #1194, opened 2026-08-18
Reintroduced ability to download audio file from article link

omck:
## What does this change?

Long pressing the "Open..." link in the article opens the audio link in the browser or whatever default is set for the system.

This ability was lost with #1172 afaik , and I wanted to keep the ability to download the files locally (The new audio player is unable to keep playing if the application loses focus on my device).

## Checklist

- [X] Ran `./gradlew ktlintFormat` and committed the result


omck:
Didn't see #1192, closing

spacecowboy:
Thanks for the PR though!

PR #1208, opened 2026-08-31
Add support for API 24+

syphyr:
## What does this change?

Updating to conscrypt 2.6.3 fixes all the previous errors that caused legacy support to be removed.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
       # Then start the emulator and run the Instrumentation tests!
       - android/start-emulator-and-run-tests:
           test-command: ./gradlew connectedDebugAndroidTest --parallel
-          system-image: system-images;android-29;google_apis;x86
+          system-image: system-images;android-25;google_apis;x86
 
       # And finally run the release build
       #- run:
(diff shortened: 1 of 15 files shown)


spacecowboy:
Thanks for the PR @syphyr . How certain are you that this version will work better? What have you tested?

syphyr:
> Thanks for the PR @syphyr . How certain are you that this version will work better? What have you tested?

Support for API 24+ in Feeder was previously removed because the older version of conscrypt being used was causing a crash. I tested over with a newer version of conscrypt and everything works as expected again. Also, using conscrypt library again would make it easier to fix crypto issues for all versions of android. The other benefit is that older legacy devices could be supported again as well.

syphyr:
small cleanup

spacecowboy:
Thank you but I no thanks. API 24 is very old by now. I do not see the value in preventing the use of modern code to support the extremely few users on that version.

And dropping Conscrypt has also made it much easier to upgrade libraries and frameworks in the app since you are not limited to keeping support with Conscrypt, which has a very strange release cycle where some releases just flat out stop working.

PR #1212, opened 2026-09-02
fix: removed prebuilt Bergamot runtime to restore F-Droid builds

JaredTweed:
## What does this change?

Removes the Bergamot fallback introduced by #1147, including its vendored WebAssembly runtime. The F-Droid source scanner rejects `app/src/main/assets/bergamot/worker/bergamot-translator-worker.wasm` before Gradle runs, which blocks the 2.23.0 update in fdroiddata!47282.

The existing Android system on-device translation provider remains available. Features merged after #1147, including the single-column feed option, built-in audio setting, and link blocklist setting, are preserved. Bergamot can be reintroduced separately once its WASM runtime is reproducibly built from source in an F-Droid-compatible toolchain.

## Verification

- `./gradlew ktlintFormat`
- `./gradlew ktlintCheck`
- `./gradlew compileFdroidDebugKotlin`
- `./gradlew lintFdroidDebug`
- `./gradlew compileFdroidDebugAndroidTestKotlin`
- `./gradlew testFdroidDebug` (305 tests passed, 3 skipped)
- `./gradlew assembleFdroidRelease`
- Confirmed no `.wasm`, `.jar`, `.aar`, `.dex`, or `.so` files remain under `app/src`

No Room schema changes.


JaredTweed:
Closing this removal-only fix in favor of preserving offline translation with an F-Droid-compliant, explicitly consented, SHA-verified runtime download. I will link the replacement PR when it is ready.

PR #1213, opened 2026-09-02
fix: moved Bergamot runtime to a verified opt-in download

JaredTweed:
## Summary

- removed the prebuilt Bergamot WebAssembly binary from the source tree and APK
- downloads the pinned `@browsermt/bergamot-translator` 0.4.5 runtime only after explicit user consent
- warns that the executable runtime is not built or checked by F-Droid
- verifies the runtime against its pinned SHA-256 before storing or loading it
- preserves offline Bergamot translation after the one-time runtime download
- avoids prompting when no Bergamot model path exists for the requested language pair

This addresses the F-Droid scanner failure without removing the Bergamot fallback. The release APK contains the JavaScript worker sources but no `.wasm` asset.

## Verification

- `./gradlew ktlintCheck test compileFdroidDebugAndroidTestKotlin installFdroidDebug`
- `./gradlew assembleFdroidRelease lintFdroidDebug`
- focused runtime tests cover successful installation, SHA mismatch rejection, and reuse of an already verified runtime
- inspected both debug and release APK contents and confirmed there are no `.wasm` entries
- installed successfully on the attached Pixel 8 Pro


spacecowboy:
Thanks @JaredTweed . I think this will be a good approach

JaredTweed:
@spacecowboy Okay, the code is ready to be merged!

spacecowboy:
Sorry for the delay. Thanks again @JaredTweed

PR #1217, opened 2026-09-10
feat: added per-feed entry filtering rules

antoncarrot:
## What does this change?

Adds **per-feed entry filtering rules**, modelled on [Miniflux's filtering rules](https://miniflux.app/docs/rules.html#filtering-rules). Each feed gets two multi-line fields on the Add/Edit Feed screen — block rules and allow rules — holding one `FieldName=regex` rule per line.

Motivation: Feeder's only content filter today is the global blocklist, which applies to *every* feed at once and matches titles with globs. That doesn't cover the common case of one noisy feed — a high-volume aggregator where you want a handful of tags or a sponsored-post pattern gone, without touching your other subscriptions.

Rules are evaluated **at fetch time**, so a matching item never becomes a row, a blob file, or a notification.

### Semantics

Supported fields: `EntryTitle`, `EntryURL`, `EntryAuthor`, `EntryTag`, `EntryContent`.

1. Block rules run first — first match wins, item dropped.
2. Then, if any allow rules are set, the item must match at least one; otherwise it is dropped.
3. Empty allow rules means "keep everything not blocked".
4. Invalid rules are ignored, never fatal.

Patterns are Java regexes matched as substrings (`(?i)` for case-insensitivity), so `EntryTitle=(?i)(sponsored|promoted)` works as you'd expect.

### How this relates to the existing blocklist

The global blocklist is untouched. The two features are deliberately different, and the UI copy says so:

| | Global blocklist (existing) | Per-feed rules (new) |
|---|---|---|
| Syntax | glob | Java regex |
| Scope | all feeds | one feed |
| Mechanism | row inserted, `block_time` stamped | row never inserted |
| Reversible | yes — clear the pattern, items reappear | **no** — gone unless the feed still serves it |
| Blob file | written | not written |

I considered stamping `block_time` instead of dropping, to get reversibility for free. Rejected: an audit of every `feed_items` consumer found four paths that don't filter `block_time` — blocked guids are still sent to the sync server once anything marks them read, blocked-but-bookmarked items still appear in saved-articles export, they stay openable by ID, and `duplicationExists` would start suppressing legitimate duplicates in *other* feeds. Stamped items would also keep their blob files and consume retention slots.

### Three deliberate deviations from Miniflux

Each of these is a case where Miniflux's behaviour silently empties a feed or fires on things the user can't see:

- **An empty pattern is rejected.** `EntryTitle=` compiles fine in Miniflux and matches *everything*, which would silently swallow an entire feed. Here it's a validation error.
- **`EntryContent` matches plain text**, falling back to the HTML only when the plain text is blank. Miniflux matches sanitized HTML, which on a phone means `EntryContent=crypto` fires on `<a href="https://crypto.example/…">` and on class names — undebuggable for the user.
- **An `EntryTag` allow rule is skipped for articles with no tags.** Matching iterates the tag list, so an untagged entry matches no `EntryTag` rule — not even `.*`. As a block rule that's harmless, but as an allow rule it means `EntryTag=(?i)linux` on a feed that emits no `<category>` drops *every* article. Allow rules are now filtered to those that can actually be evaluated against the article before a match is required; if none apply, the article is kept, exactly as if no allow rules were set. Allow rules on other fields are unaffected, so `EntryTag=(?i)linux` + `EntryTitle=(?i)kernel` still filters untagged articles by title. Block rules are unchanged.

### Failing open

Invalid rules are dropped at parse time, in **both** directions. If every block rule is invalid, nothing is blocked. If every allow rule is invalid, everything is *kept* — not everything dropped. Both directions fail toward keeping data, which is the only safe direction for a filter that drops items permanently. There's an explicit test for each; a naive implementation that tracked "user typed allow rules" separately from "allow rules parsed" would silently discard the whole feed.

Invalid rules warn but do **not** block saving. A regex is invalid *while being typed* (`(?i)(foo` on the way to `(?i)(foo|bar)`), so a flickering-disabled OK button would hold unrelated title/tag/notify edits hostage over a typo. The error is never silent: red field, red label, and a line naming the exact line number.

### Schema, OPML and sync

- Schema **39 → 40**: `feeds.block_rules` and `feeds.allow_rules`, both `TEXT NOT NULL DEFAULT ''`. `MIGRATION_39_40` plus a migration test; generated `40.json` committed.
- Rules round-trip through OPML. Note XML normalizes whitespace in attribute values, so newlines and tabs are written as `&#10;` / `&#9;` — otherwise a multi-line rule comes back as one corrupted line. Only the two new attributes go through the new escaping, so every existing exported byte is unchanged.
- Rules round-trip through Feeder Sync. As with every previously added field, a peer on an older build that edits the feed sends `""` and, if its `whenModified` is newer, wipes the rules.

### Known limitations

- **Future fetches only.** Adding a rule does not remove articles you already have, and removing a rule does not bring back what was skipped.
- **ReDoS is possible in principle.** Miniflux can accept arbitrary user regexes safely because RE2 is linear-time; `java.util.regex` backtracks with no timeout. A pattern like `EntryContent=(a+)+$` against a large body would spin inside `singleThreadedSync`. Bounded by a 100 000-character match cap and a 100-rule cap, but not hard-guaranteed. Accepted: the pattern is authored by the device owner and affects only their device.
- **Pruning edge case.** Filtered items are absent from `presentIds`, so if a feed shrinks (or you lower max-items-per-feed) such that stored rows exceed `keepCount`, a still-served article that the rules now filter can be pruned. `keepCount` is `max(maxFeedItemCount, items.size)` over the *raw* parsed list, so this can't happen at steady state. Documented in a comment next to `keepCount`; an earlier guard against it was reverted as not worth making every filtered item hit the database twice.

### Drive-by fix

`CreateFeedScreenViewModel` was silently dropping `summarizeOnOpen` when creating a feed.

### Testing

- 41 JVM tests in `EntryFilterRulesTest` covering parsing, matching, and policy — this is where the feature is specified.
- 6 instrumented tests in `RssLocalSyncKtTest`, including `blockedItemsGetNoBlobFile` (proves the drop happens before `upsertFeedItems`) and `itemsStoredBeforeRuleWasAddedAreNotDeleted`.
- `TestMigrationFrom39To40`, an OPML round-trip test through the real `XmlPullParser`, and `EncryptedFeed` round-trip tests.
- Full suite green on a Pixel emulator: `./gradlew check` clean, 202 instrumented tests, 0 failures.

Also verified by hand on a device, since two things don't show up in tests: the keyboard's return key inserts a newline rather than dismissing (`ImeAction.Default` — with `singleLine = false`, using `Done` here would make a second rule impossible to type), and the inline validation renders correctly with the OK button still enabled.

## Checklist

## Screenshots (if UI change)

| The new section on Edit Feed | Inline validation |
|---|---|
| s4 | s6 |


app/src/main/java/com/nononsenseapps/feeder/model/EntryFilterRules.kt
```diff
@@ -0,0 +1,259 @@
...
+data class EntryRuleSet(
+    val rules: List<EntryRule>,
+    val errors: List<EntryRuleError>,
+) {
+    companion object {
+        val EMPTY = EntryRuleSet(emptyList(), emptyList())
+    }
+}
+
+object EntryRules {
```
spacecowboy: Why is this an Object?

The `parse` function could just be a plain function.

`MAX_RULES` could just be a const in the file level as well.

spacecowboy (changes requested):
Thanks for the PR. Overall I think it looks good. I have left an inline comment.

The biggest comment I have is that nothing should be called `Entry`. It would align much better with the terminology the App uses if `Entry` was renamed with `Article` everwhere.

But, it would also just be simpler if for example `EntryTitle` was just called `Title` (and not `ArticleTitle`). The rules are all applied to Articles. There is no doubt which `Title` the filter is referring to. This is specifically the user facing strings. It is preferable fine if the internal data classes etc are called `Article...` etc.

tldr:

* user facing strings: remove mention of `Entry`. Just `Title` and `Tags` etc are enough.
* code: Rename `Entry` to `Article`
* See inline comment about `Object`

app/src/main/java/com/nononsenseapps/feeder/model/EntryFilterRules.kt
antoncarrot: move parse() onto EntryRuleSet

antoncarrot:
> user facing strings: remove mention of Entry. Just Title and Tags etc are enough.

done

> code: Rename Entry to Article

done

> See inline comment about Object

refactored, moved parse() onto EntryRuleSet

spacecowboy:
Thanks @antoncarrot . That's much easier to read I think.

spacecowboy (approved):
I think this is OK to test now

